PR #2, opened 2017-06-19
Fix README links

martinthomson:
I wonder if this isn't a consequence of an editor configuration issue, maybe in @mnot's editor, because there was definitely no spaces in the script that generated this file.


britram:
this was my editor, my bad... need to hack vs code's reflow stuff to be nice to links.

PR #4, opened 2017-06-21
Cleanup

martinthomson:
This brings the ops drafts into line with the base ones.  The configuration (circle.yml, Makefile, .editorconfig) is copied over.

I removed trailing whitespace from the drafts and reflowed paragraphs.

I changed the references so that they are a little cleaner.  Other than the reflow, I didn't touch the text, other than to change "section" to "Section" where I noticed some errors.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# See http://editorconfig.org
+
+root = true
+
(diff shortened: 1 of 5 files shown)


britram:
thanks!

PR #21, opened 2017-11-01
add application identification section

britram:


draft-ietf-quic-manageability.md
```diff
@@ -284,12 +284,24 @@ future versions of the protocol.
 
 ## Connection confirmation {#sec-confirm}
 
-Connection establishment requires cleartext packets and is using a TLS handshake
-on stream 0. Therefore it is detectable using heuristics similar to those used
-to detect TLS over TCP. 0-RTT connection may additional also send data packets,
-right after the Client Initial with the TLS client hello. These data may be
-reordered in the network, therefore it may be possible that 0-RTT Protected data
-packets are seen before the Client Initial packet.
+Connection establishment uses Client Initial, Server Cleartext, and Client
```
martinthomson: We just changed the name of these packets.  Sorry.

draft-ietf-quic-manageability.md
```diff
@@ -284,12 +284,24 @@ future versions of the protocol.
...
+
+## Application Identification {#sec-server}
+
+The cleartext TLS handshake may contain Server Name Indication (SNI)
+{{?RFC6066}}, by which the client reveals the name of the server it intends to
+connect to, in order to allow the server to present a certificate based on
+that name. It may also contain information from Application-Layer Protocol
+Negotiation (ALPN) {{?RFC7301}}, by which the client exposes the names of
+application-layer protocols it supports and the server exposes the name of the
+selected protocol.
```
martinthomson: This isn't accurate.  The client list of ALPN options is in the clear, but the server choice from that list is not.

draft-ietf-quic-manageability.md
britram: NP, was tracking -07. 

The change also changes how handshake tracking works, which I'll address in a separate PR.

britram:
Re SNI and ALPN: we can certainly point at the work in progress in TLS that will encrypt this information, but IMO the document should describe the current state of the world.

mirjak:
I would actually prefer to describe the wire image of TLS and associated protocols in a separate document and just point to it because that's actually not only applicable for quic

martinthomson:
I don't have a reference for encrypting ALPN.  We don't have one.  Maybe I should write one...  On the plus side, it's something that benefits from the same techniques as SNI (and less likely to cause deployment issues, though that is only because we haven't ossified around it as much).

As for the general point about now vs. later.  One of the things we've tried to do is be very clear about what the invariants are, or what promises we are making.  It's not the same thing as true ossification defense, but if we make it clear what we are committing to, maybe we won't have people relying on stuff we might want to change.

PR #24, opened 2018-01-09
Use of the Latency Spin Bit

britram:
This PR takes text from [draft-trammell-quic-spin](https://tools.ietf.org/html/draft-trammell-quic-spin) to explain the on-path use of the latency spin bit. It is the manageability draft companion of the new [latency spin bit PR](https://github.com/quicwg/base-drafts/pull/1046).

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -351,21 +351,22 @@ Handshake RTT can be measured by adding the client-to-observer and
 observer-to-server RTT components together. This measurement necessarily
 includes any transport and application layer delay at both sides.
 
-RTT measurements during a flow are are possible by observing the series of the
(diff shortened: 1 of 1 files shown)


larseggert:
I don't think we want to merge this until we have consensus that the spin bit itself is merged?

britram:
fbc8490 changes this to point to the experiment (quic-spin-exp), not to the protocol.

mirjak:
To further clarify, the draft was mentioning that there is a discussion about the spin bit before. This PR mainly only now points to the respective draft. I think it is far to have this pointer in there for now.

larseggert:
Well, as long as you are prepared to possibly remove this, fine

britram:
The nice thing about Internet-Drafts is that they are endlessly mutable. :)

PR #45, opened 2018-10-18
Timing and CID linkability

britram:
Add a section about being careful with timing with respect to CID migration if you actually care about linkability.

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -346,13 +346,13 @@ a link among these addresses to observers on the network. However, analysis of
 the lifetimes of six-tuples (source and destination addresses as well as the
 migrated CID) may expose these links anyway.
 
-The key insight here is that a given observation point in the network, a finite
-set of flows will be undergoing migration within any one time window, and any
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-applicability.md
```diff
@@ -336,6 +336,30 @@ forward the traffic based on the Connection ID. A server can choose the
...
+
+QUIC provides for multiple Server and Client Connection IDs to be used
+simultaneously by a given connection, which allows seamless connection migration
+when one of the endpoints changes IP address and/or UDP port. Section 6.11.5 of
+{{QUIC}} describes how to use this facility to reduce the risk of exposing
+a link among these addresses to observers on the network. However, analysis of
+the lifetimes of six-tuples (source and destination addresses as well as the
+migrated CID) may expose these links anyway.
+
+The key insight here is that a given observation point in the network, a finite
```
chris-wood: Perhaps: In practice, a finite set of flows will be undergoing migration within any one time window from the perspective of a given observation point in the network. A key insight here is that any migration must keep at least one endpoint address constant during the migration. This finite set...

The key insight here is that , a finite

PR #47, opened 2018-10-19
move ConnID generation language to -applicability

britram:
-manageablility has guidelines on how to select ConnIDs at a server, which is really more an -applicability thing. This fixes that.

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -299,6 +299,18 @@ not behind a load balancer, and therefore it is unlikely that the addresses
 will be re-bound, may indicate to the transport that is wishes to not expose a
 Connection ID.
 
+Server-generated Connection IDs must not encode any information other that that
+needed to route packets to the appropriate backend server(s): typically the
+identity of the backend server or pool of servers, if the data-center’s load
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-applicability.md
```diff
@@ -299,6 +299,18 @@ not behind a load balancer, and therefore it is unlikely that the addresses
 will be re-bound, may indicate to the transport that is wishes to not expose a
 Connection ID.
 
+Server-generated Connection IDs must not encode any information other that that
```
nibanks:
```suggestion
Server-generated Connection IDs must not encode any information other than that
```

draft-ietf-quic-applicability.md
```diff
@@ -299,6 +299,18 @@ not behind a load balancer, and therefore it is unlikely that the addresses
 will be re-bound, may indicate to the transport that is wishes to not expose a
 Connection ID.
 
+Server-generated Connection IDs must not encode any information other than that
```
nibanks: Hey, wait... this entire blurb is already in the following section it looks like. Right?

britram: oh, indeed. we already copied this over, but didn't remove it from -manageability.

PR #41, opened 2018-10-17
Text to section on keep-alives vs. resumption

mirjak:
address issue #6

PR #50, opened 2018-10-19
Keepalives versus 0RTT, remix

britram:
An alternate approach to #41, following from @martinthomson's comment thereon.


mirjak:
I think the new text is now missing two points: 
1) It might make sense to send keep-alives if the application know that is there is new data coming, e.g. an outstating reply or a following request. I think this is what is recommended for http.
2) as meantion in the transport draft we should probably say again that it must be clearly specified which party is performing the active ping-ing

Also the text basically "just" explains the situation but doesn't really make any recommendation. Do we maybe what to say more than that?

martinthomson:
Yay, unicorns mean that I can't see @mirjak's comments here, though they appeared in email.

In response to those, maybe it is OK to point to HTTP/QUIC's recommendations about keep-alives.  There, it recommends limiting use of PING for keep-alive to while a request is outstanding.

PR #52, opened 2018-10-26
Upgrade and rollback processes

### LINKED ISSUE #28: Document how a server cluster might be upgraded and rolled back
Moving this from quicwg/base-drafts#504 to here.

martinthomson:
@kazuho should probably double-check my work here.

This assumes that we take the compatible version upgrade stuff, but if we don't, there's only one sentence to remove.  I've tried to keep it abstract enough that the details won't matter.

Closes #28.


kazuho:
LGTM. Thank you for writing it down!

PR #55, opened 2018-11-14
For privacy concerns, the spin bit may be disabled in specific use cases.

loganaden:


loganaden:
ping

britram:
pong, back from leave, looking at these drafts this week...

britram:
The document definitely needs to be updated to reflect this decision from BKK (see #58), but I think the reasons here are more comprehensive than just proxies.

loganaden:
Thank you for your feedback. I will update the proposed patch.

britram:
AFAICT this has been overtaken by changes to the spin bit text in -transport as well as by #66

PR #82, opened 2019-07-04
Make it clear in the intro that we don't do proxies here.

### LINKED ISSUE #76: Document the challenges of creating a 'transparent' QUIC MITM
Feel free to close this if it's out of scope, but creating a proxy that attempts to passively observe QUIC is much more complex and potentially problematic than TLS over TCP.

For example, all major OS's have mature TCP implementations, whereas a MITM would have to pull in its own QUIC implementation.  If it does so, it would be important that's it's both conformant to IETF specifications and have good performance.  This is the full L7 termination case, which is possible, but a lot of work, and we should heavily encourage anyone doing this to use an existing("good") implementation.

A potential MITM option is to attempt to terminate the handshake, but then only observe ApplicationData packets.  However, this could go wrong in many ways, so I believe it should be strongly discouraged.  QUIC is designed as an end-to-end protocol and efforts to change that are likely to end very poorly.

For example, if the length of the CID changes, then the offset of the header changes and a middlebox who doesn't track all CIDs both peers may be unable to decrypt packets.  I'm sure there are more gotchas, particularly with the introduction of extensions.  It may also be worth pointing out that QUIC frames are not TLV, so any frames(ie: extensions) that are not understood cause all other frames to be unparseable.

Thanks to @DavidSchinazi for the specific examples of ways this could fail.

britram:


draft-ietf-quic-manageability.md
```diff
@@ -94,6 +94,12 @@ assumptions that the QUIC protocol design takes toward the expected network
 treatment. It also discusses how common network management practices will be
 impacted by QUIC.
 
+Since QUIC's wire image {{?WIRE-IMAGE=RFC8546}} is integrity protected and not
+modifiable on path, in-network operations are not possible without terminating
+the QUIC connection, for instance using a back-to-back proxy. Proxy operations
+are not in scope for this document, as QUIC proxies must be fully-fledged QUIC
+endpoints.
```
mirjak: Maybe add another half sentence saying something like "and therefore should follow guidance provided in the QUIC base specs {{...}} and the applicability statement {{...}}"...?

britram: yep

Issue #87, opened 2019-11-16
Considerations for NATs


martinduke:
The manageability draft should talk a little bit about NATS for QUIC.

The laziest thing to do is just continue to have NAT mapping be by 4-tuple, as it is today. This is actually fine.

However, someone might attempt to be clever and use the CID as well. For example, client A and client B could share the same serverside address and port and the NAT can initially distinguish between the two because they have different CIDs. This appears to save address space,  but doesn't actually work because the server could switch to a new client connection ID. In this case, the NAT won't be able to determine which client to route to.

erickinnear:
I think it's also worth mentioning that using just the CID only is similarly broken, the fundamental underlying concept here is that subsequent CIDs are exchanged in a manner that the NAT cannot see -- the only time it's okay to use CIDs is if you're colluding with one of the endpoints, as in QUIC LB.

larseggert:
At least recommend to use longer binding lifetimes for identified QUIC connections?

mirjak:
See PR #88

britram:
Fixed in this doc by #88

PR #88, opened 2019-12-09
No CIDs for NATs, addresses #87

britram:


draft-ietf-quic-manageability.md
```diff
@@ -696,7 +696,11 @@ requirements.
 The QUIC header optionally contains a Connection ID which can be used as
 additional entropy beyond the 5-tuple, if needed. The QUIC handshake needs
 to be observed in order to understand whether the Connection ID is present and
-what length it has.
+what length it has. However, Connection IDs may be renegotiated during
```
nibanks: Perhaps it should be explicitly mentioned that even the length of the CIDs can change. A change in length could break any logic that observes a handshake and assumes all CIDs have the same length.

Not to mention that the tuple itself could be shared by several parallel connections, each with different CIDs (possibly with different lengths).

martinduke:
At Gorry's suggestion I put the NAT guidance into a different draft:
https://datatracker.ietf.org/doc/draft-duke-quic-natsupp/

Sorry, I thought the notification went out automatically. Regardless, this change is fine.

draft-ietf-quic-manageability.md
```diff
@@ -696,7 +696,11 @@ requirements.
...
+a connection, and this renegotiation is not visible to the path. Keying state 
+off the Connection ID may therefore cause undetectable and unrecoverable loss 
+of state in the middle of a connection. Use of Connection ID specifically
```
ianswett:
```suggestion
of state in the middle of a connection. Use of Connection ID is specifically 
```

ianswett: Or maybe you want the is below, but I think something is missing?

Issue #93, opened 2020-04-27
Improve and generalize text on DoS detection and mitigation


ximaera:
> Current practices in detection and mitigation of Distributed Denial of Service (DDoS) attacks generally involve passive measurement using network flow data {{?RFC7011}}, classification of traffic into "good" (productive) and "bad" (DoS) flows, and filtering of these bad flows in a "scrubbing" environment

This is not how it works. A few examples:
 - https://en.wikipedia.org/wiki/SYN_cookies
 - https://www.researchgate.net/publication/220027076_Mitigating_denial_of_service_attack_using_CAPTCHA_mechanism

britram:
I'm not sure I see support for "this is not how it works" in your examples: SYN cookies on their own are a 1990s technique for end-system DoS prevention, which, while useful at transport-terminating front ends as part of a larger DoS prevention strategy, aren't really usable on their own at scale. The paper you link to here (from ICWET, and itself a decade old) primarily deals with the separation of good from bad flows at layer 7, not necessarily how the front-end handles these flows once classified. But you're definitely right that this is not *only* how it works: the DoS mitigation landscape is wide and varied (even if at the end of the day the only strategy we have found to work at scale is "eat the DoS".

That said, this text could be improved: it discounts detection and classification mechanisms other than based on flow designators derived from passive measurement, does not mention protocols other than IPFIX and the flow model that comes along with it when classifying traffic, etc. 

The text should also be more applicable to UDP reflection DoS, which is a primary concern in wire image differentiation in QUIC (and which also isn't otherwise addressed in the document or in this issue). Adding a path-observable SYN cookie equivalent to QUIC, while useful, seems difficult to achieve given the working group's history of consensus about path-observable wire image features, so other classification mechanisms are what we'll be left with.

ximaera:
> it discounts detection and classification mechanisms
> other than based on flow designators derived from
> passive measurement

Exactly my point.

The passive measurements are generally only useful to prevent the collateral damage to the network infrastructure of an ISP or a datacenter network under an attack. While the infrastructure impact of QUIC is itself an intriguing research topic, DDoS affects not only transit networks but also the endpoints in ways that could not be prevented with any passive flow classification.

britram:
[...]

PR #94, opened 2020-05-29
DDoS mitigations - linking after NAT rebinding and UDP Policing

igorlord:
This PR describes how a stateful DDoS system may link flows after a NAT rebinding event.

Also, the PR adds a section on how to police QUIC packets to minimize disruption to application that are able to failover to TCP.  Namely, avoid blocking a random fraction of UDP packets and, instead, focus on either blocking all packets from a subset of users or none at all.

This is related to areas discussed in #93.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -744,6 +744,23 @@ performs the DDoS detection, an active, migrated QUIC connection may be
 blocked by such a system under attack. However, a defense system might simply
 rely on the fast resumption mechanism provided by QUIC.
 
+### UDP Policing
+
+UDP is the most prevalent DDoS vector, since it is easy for compromised
+non-admin applications to send a flood of large UDP packets (while with TCP the
+attacker gets throttled by the congestion controller). Networks should be
+prepared for UDP flood attacks that masquerade as QUIC traffic.
+
+If attack classification fails, a possible response is to police UDP traffic on
+the network, allocating a fixed portion of the network capacity to UDP. Blindly
+blocking a significant fraction of QUIC packets will allow many QUIC handshakes
+to complete, preventing a TCP failover, but the connections will suffer a severe
+packet loss. As long as all QUIC-capable applications can failover to TCP (at
+least applications using well-known UDP ports), the recommended way to drop QUIC
+packets is to either drop them all or to police them based on the hash of the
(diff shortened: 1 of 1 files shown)


nibanks:
There's no mention of the QUIC Retry mechanism here. At the very least, for long header packets, a cooperating DDoS protection device can maintain a white-list of peer IP addresses, and only let through those addresses that successfully go through the Retry scenario.

The question then is what to do with short header packets. The backend server will not have any connection state for any junk short header packets, so the server should either quickly discard the packet silently or send a stateless reset if it happens to look like a valid QUIC packet. Assuming the short header packet traffic is enough to overwhelm the server, then this PR's text would come into play. Though it'd be interesting to see some perf numbers on what kind of CPU cost there is to just reject these packets. From my previous tests with the QUIC server running in the Windows kernel, the network was the limiting factor in these type of attacks, not the CPU. So if the network is the bottleneck, does it really matter if you drop it at the server or just in front of it?

igorlord:
@nibanks I think this section could be expanded a lot more, and Retry mechanism with a cooperating DDoS protection device definitely has a place. I am happy to give it a try in a different PR, or I can leave it to you, since you've likely thought more about this topic. The cooperation should include an agreement on the format of the Token to:
1. Determine whether the Token is a Rety token or a token provided by a NEW_TOKEN frame.
1. For a Retry token, decode original_destination_connection_id.

As for the CPU cost, this consideration only comes into play, if the attack was allowed to reach backend servers.  Volumetric DDoS can saturate network links and/or be targeting networking infrastructure and not just backend servers.  As @britram is saying in #93, volumetric attacks are often mitigated further away from the edge of the network.  This kind of mitigation is the focus of UDP Policing section.

igorlord:
looks like https://tools.ietf.org/html/draft-ietf-quic-load-balancers-02#section-5 is an excellent discussion of Retry mechanics.  It definitely deserves a reference from this doc.

The Retry mechanisms described in LB draft are all cooperating boxes, and servers are aware of them.

Unfortunately, Retry system protects only server's memory state and some CPU cycles spent on crypto.  (Servers still need to decrypt CID to decide it is invalid, and if the attacker is clever enough to establish one valid connection and use that CID in a flood, the server will also be decrypting packets.)  Retry does nothing to protect network resources.

This PR is about transparent boxes that try to mitigate volumetric attacks.

PR #99, opened 2020-07-06
Edits on UDP policing section

mirjak:


draft-ietf-quic-manageability.md
```diff
@@ -750,22 +750,26 @@ client, the client may rely on the fast resumption mechanism provided by
...
+prepared for UDP flood attacks on ports used for QUIC traffic. One possible
+response to this threat is to police UDP traffic on the network, allocating a
+fixed portion of the network capacity to UDP and blocking UDP datagram over that
+cap.
+
+The recommended way to police QUIC
+packets is to either drop them all or to throttle them based on the hash of the
 UDP datagram's source and destination addresses, blocking a portion of the hash
 space that corresponds to the fraction of UDP traffic one wishes to drop.
+When then handshake is blocked, QUIC-capable applications may failover to TCP
```
martinthomson, line 768: "When then"

martinthomson, line 768: "may" only really applies if there is a path to TCP.  If you have addressed that point elsewhere, fine, but otherwise you need to say something about that.

draft-ietf-quic-manageability.md
```diff
@@ -750,22 +750,26 @@ client, the client may rely on the fast resumption mechanism provided by
...
+(at least applications using well-known UDP ports). However, blindly blocking a
+significant fraction of QUIC packets will allow many QUIC handshakes
+to complete, preventing a TCP failover, but the connections will suffer a severe
```
martinthomson, line 771: "a severe" might be better as "from severe" or just "severe"

draft-ietf-quic-manageability.md
mirjak: done. thanks!

draft-ietf-quic-manageability.md
mirjak: done

draft-ietf-quic-manageability.md
mirjak: What do you mean by "path to TCP"? Doesn't "may" already include that is may also not work (for whatever reason)? Should we just say "may try to failover"?

martinthomson: This usage of "may" implies that this is an option.  As in "can". As does your "try to". This is more about the various ways in which failures are handled. Even in HTTP, we don't fall back, we just fail to upgrade to QUIC.  Implying that callbacks are involved might be a convenient colloquialism, but it isn't accurate. And then there are those protocols that might come to only use QUIC, which will just fail.

PR #102, opened 2020-07-07
More guidance on idle timeouts

mirjak:


draft-ietf-quic-applicability.md
```diff
@@ -452,12 +452,20 @@ human-readable information to an error code. Note that QUIC RESET_STREAM and
...
-application. If an application desires to keep the connection open for longer
-than the announced timeout, it can send keep-alives messages. See
-{#resumption-v-keepalive} for further guidance.
+Alternatively, a QUIC connection can be silently closed by each endpoint
+separately after an idle timeout. If enabled as indcated by a transport
+parameter in the handshake, the idle timeout is announced for each
+endpoint during connection establishment and the effective value for this
+connection is the minimum of these two advertised value. An application
+therefore should be able to configure its own maximum value as well as
+have access to the computed minimum value for this connection. An application
```
marten-seemann, line 461: I don’t understand why an implementation needs access to the computed value. A reasonable QUIC implementation could take the application’s idle timeout and then take care of sending PINGs to the peer to keep the connection open.

draft-ietf-quic-applicability.md
```diff
@@ -452,12 +452,20 @@ human-readable information to an error code. Note that QUIC RESET_STREAM and
...
+may adjust the maximum idle timeout based on the number of open or expected
+connections as shorter timeout values may free-up memory more quickly.
+If an application desires to keep the connection open for longer
+than the announced timeout, it can send keep-alives messages, or a QUIC
+implementation may provide an option to defer the time-out to avoid unnesseary
```
marten-seemann, line 466: Not every application protocol has a keep-alive option. I view keep-alives as the responsibility of the QUIC stack instead.

draft-ietf-quic-applicability.md
britram: The text here, I think, is appropriate for general-purpose QUIC implementations (atop which multiple applications could be bound), but yeah, it could be either-or for applications tightly bound to their implementation, based on application keepalive logic.

draft-ietf-quic-applicability.md
mirjak: Both might have benefits depending on the application, e.g. if you anyway need to exchange certain status update messages on application layer. Should be more explicit about that in the text?

draft-ietf-quic-applicability.md
```diff
@@ -452,12 +452,20 @@ human-readable information to an error code. Note that QUIC RESET_STREAM and
...
 are defined to be applicabile to all three frames.
 
-Alternatively, a QUIC connection will be silently closed by each endpoint
-separately after an idle timeout. The idle timeout is announce for each
-endpoint during connection established and should be accessible by the
-application. If an application desires to keep the connection open for longer
-than the announced timeout, it can send keep-alives messages. See
-{#resumption-v-keepalive} for further guidance.
+Alternatively, a QUIC connection can be silently closed by each endpoint
+separately after an idle timeout. If enabled as indcated by a transport
```
martinthomson, line 456:
```suggestion
separately after an idle timeout. If enabled as indicated by a transport
```

draft-ietf-quic-applicability.md
```diff
@@ -452,12 +452,20 @@ human-readable information to an error code. Note that QUIC RESET_STREAM and
...
+parameter in the handshake, the idle timeout is announced for each
+endpoint during connection establishment and the effective value for this
+connection is the minimum of these two advertised value. An application
```
martinthomson, line 459:
```suggestion
connection is the minimum of the two advertised values. An application
```

draft-ietf-quic-applicability.md
martinthomson, line 466:
```suggestion
implementation may provide an option to defer the time-out to avoid unnecessary
```

draft-ietf-quic-applicability.md
```diff
@@ -452,12 +452,20 @@ human-readable information to an error code. Note that QUIC RESET_STREAM and
...
+therefore should be able to configure its own maximum value as well as
+have access to the computed minimum value for this connection. An application
+may adjust the maximum idle timeout based on the number of open or expected
+connections as shorter timeout values may free-up memory more quickly.
+If an application desires to keep the connection open for longer
+than the announced timeout, it can send keep-alives messages, or a QUIC
+implementation may provide an option to defer the time-out to avoid unnesseary
+load, as specified in Section 10.2.2 of {{QUIC}}.
```
martinthomson, line 467: Be aware that this anchor is likely to change to 10.1.2.

draft-ietf-quic-applicability.md
```diff
@@ -452,12 +452,20 @@ human-readable information to an error code. Note that QUIC RESET_STREAM and
...
+See {#resumption-v-keepalive} for further guidance on keep-alives.
```
martinthomson, line 468:
```suggestion
See {{resumption-v-keepalive}} for further guidance on keep-alives.
```

draft-ietf-quic-applicability.md
martinthomson: I think that - at the level of this document - this is fine advice.  Applications do benefit from knowing when a timeout is imminent because they might want to manage the risk of failure for transactions that can't be easily retried.  At or near the timeout is not a great time to start something.

pravb:
One concern I have with short idle timeout is impact on VM live migration. The current guidance with TCP default timeouts allows LM to have a blackout period of under 9 seconds. Do we need any guidance here to not use very low default timeouts?

PR #108, opened 2020-07-07
More improvements to QUIC identification

### LINKED ISSUE #96: Mention accidental invariants and the calculus of blocking
From Spencer on the list:

> Before I read this ^ paragraph, I was planning to ask you whether "blocking QUIC" merited a mention in https://tools.ietf.org/html/draft-ietf-quic-manageability-06, because we are in the awkward position with HTTP/3 that operators will face less blowback blocking HTTP/3-QUIC and forcing a fallback to HTTP/2-TCP than they will face blocking almost any non-HTTP/3 QUIC protocol (with less clearly defined fallbacks) (see: MASQUE). 
> ISTM that at a minimum, the "accidental invariants" might be mentioned in https://tools.ietf.org/html/draft-ietf-quic-manageability-06, which targets operators.

britram:
Fixes #96

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -493,18 +493,23 @@ carried on the wire) unless noted.
 The QUIC wire image is not specifically designed to be distinguishable from
 other UDP traffic.
 
-The only application binding currently defined for QUIC is HTTP
-{{?QUIC-HTTP}}. HTTP over QUIC uses UDP port 443 by default,
+The only application binding defined by the IETF QUIC WG is HTTP/3
+{{?QUIC-HTTP}} at the time of this writing; however, work to run other
+applications over QUIC is ongoing, so an assumption that all QUIC traffic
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -493,18 +493,23 @@ carried on the wire) unless noted.
 The QUIC wire image is not specifically designed to be distinguishable from
 other UDP traffic.
 
-The only application binding currently defined for QUIC is HTTP
-{{?QUIC-HTTP}}. HTTP over QUIC uses UDP port 443 by default,
+The only application binding defined by the IETF QUIC WG is HTTP/3
+{{?QUIC-HTTP}} at the time of this writing; however, work to run other
+applications over QUIC is ongoing, so an assumption that all QUIC traffic
```
marten-seemann, line 498: Not only is the specification work ongoing, but other protocols are already being deployed in the wild (and have been for quite a while).

Issue #75, opened 2019-06-24
Document how SNI from QUIC handshakes is presented


ianswett:
SNI is currently mentioned in the Manageability document, but you'd have to read a substantial chunk of the transport and tls drafts to understand how to extract it.

I'd suggest a short section that describes how it's presented on the wire with references to the appropriate sections on transport and tls.

Currently, the section on client Initial says: "The Client Hello datagram exposes version number, source and destination connection IDs, and information in the TLS Client Hello message, including any TLS Server Name Indication (SNI) present, in the clear."

I think saying the SNI is in the clear is a bit misleading to a casual observer.

mirjak:
@ianwett can you maybe provide some text?

ianswett:
It'll likely be a large PR, but yes, I'm happy to.

mirjak:
Thanks!

nharper:
The algorithm for parsing SNI from a QUIC initial packet is specific to v1. (Future versions will change the key; they could also change the algorithm for encrypting/obfuscating initial packets, and could make other changes to the crypto handshake.) Should it be mentioned "this only works for QUICv1" with the re-written SNI bits, or somewhere else to cover more broadly "anything you do that involves decrypting packets protected by initial keys is v1 specific"?

igorlord:
> "anything you do that involves decrypting packets protected by initial keys is v1 specific"?

I think the entire draft can be prefaced by "anything you do that involves using QUIC protocol properties not explicitly specified in Invariant is v1 specific". This applies to the discussion of connection IDs, spin bit, initial keys, etc.

mirjak:
I opened a separate issue for that (#95).

martinthomson:
Note that this needs to recognize the possibility of Encrypted ClientHello.

britram:
It also needs to recognize that connection migration may break assumptions about the stability of SNI information.

ianswett:
And NAT rebinding. And mention CIDs aren't stable, particularly when the path changes.

mirjak:
another reference to potentially integrate: https://www.chromium.org/quic/parse-client-hello

martinthomson:
Having read that document, it would need to be framed to avoid condoning it.  It advocates a strategy that is not compliant with QUIC invariants.  It's very much rooted in 2020, including parsing rules for Google QUIC, so it might not be a good basis for an RFC.

PR #124, opened 2020-09-07
New section on SNI parsing and appendix on Google QUIC

mirjak:
This PR adds a new section on extracting the SNI. I also copied (most of) the pseudo code from Davids document into the appendix. We need to decide if we want to keep this and if so, potentially add some more explanation there.

Also we could further strengthen the point out that both the 5-tuple and conn ID can change (as in David's doc). There is already text directly following the new SNI section on "flow association". If we want stronger text there, maybe we can capture that in a separate issue though.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -938,6 +938,9 @@ This section contains algorithms that allows parsing versions from both
 Google QUIC and IETF QUIC. These mechanisms will become
 irrelevant when IETF QUIC is fully deployed and Google QUIC is deprecated.
 
+Note that other than this appendix, nothing in this document applies to
+Google QUIC.
+
 Conceptually, a Google QUIC version is an opaque 32bit field. When we refer to a
 version with four printable characters, we use its ASCII representation:
 for example, Q050 refers to \{'Q', '0', '5', '0'\} which is equal to
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -930,8 +930,8 @@ irrelevant when IETF QUIC is fully deployed and Google QUIC is deprecated.
 
 Conceptually, a Google QUIC version is an opaque 32bit field. When we refer to a
 version with four printable characters, we use its ASCII representation:
-for example, Q050 refers to {'Q', '0', '5', '0'} which is equal to
-{0x51, 0x30, 0x35, 0x30}. Otherwise, we use its hexadecimal representation:
+for example, Q050 refers to \{'Q', '0', '5', '0'\} which is equal to
+\{0x51, 0x30, 0x35, 0x30\}. Otherwise, we use its hexadecimal representation:
 for example, 0xff00001d refers to {0xff, 0x00, 0x00, 0x1d}.
 
 QUIC versions that start with 'Q' or 'T' followed by three digits are 
@@ -941,9 +941,9 @@ Versions Q046, Q050, T050, and T051 are not fully documented, but this appendix
 should contain enough information to allow parsing Client Hellos for those versions.
 
 This appendis uses the following conventions:
-array[i] – one byte at index i of array
-array[i:j] – subset of array starting with index i (inclusive) up to j-1 (inclusive)
-array[i:] – subset of array starting with index i (inclusive) up to the end of the array
+array\[i\] – one byte at index i of array
+array\[i:j\] – subset of array starting with index i (inclusive) up to j-1 (inclusive)
+array\[i:\] – subset of array starting with index i (inclusive) up to the end of the array
 
 ### Distinguishing IETF QUIC and Google QUIC Versions
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -589,7 +589,7 @@ Work is currently underway in the TLS working group to encrypt the SNI in TLS
 would make SNI-based application identification impossible through passive
 measurement.
 
-### Extracting the Server Name Indication (SNI) information
+### Extracting Server Name Indication (SNI) Information
 
 If the SNI is not encrypted it can be derived form the QUIC Initial packet 
 by calculating the Initial Secret to decrypt the packet payload and parse the 
@@ -603,7 +603,7 @@ the QUIC packet is set to 1, indicating a long header.
 
 Note, that proprietary QUIC versions, that have been deployed before standardization,
 might not set the first bit in a QUIC long header packets to 1. To parse these
-versions example code is provided in the appendix (see {{#sec-google-version}}),
+versions example code is provided in the appendix (see {{sec-google-version}}),
 however, it is expected that these versions will gradually disappear over time.
 
 When the version has been identified as QUIC version 1, the packet type needs to be
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -939,7 +939,7 @@ Google QUIC and IETF QUIC. These mechanisms will become
 irrelevant when IETF QUIC is fully deployed and Google QUIC is deprecated.
 
 Note that other than this appendix, nothing in this document applies to
-Google QUIC. And the purpose of this appendix is merely to distinguish IETF QUIC 
+Google QUIC. And the purpose of this appendix is merely to distinguish IETF QUIC
 from any versions of Google QUIC.
 
 Conceptually, a Google QUIC version is an opaque 32bit field. When we refer to a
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -863,11 +907,167 @@ security-relevant.
 
 Dan Druta contributed text to {{sec-ddos-dec}}. Igor Lubashev contributed text
 to {{sec-loadbalancing}} on the use of the connection ID for load balancing.
-Marcus Ilhar contributed text to {{sec-rtt}} on the use of the spin bit.
+Marcus Ilhar contributed text to {{sec-rtt}} on the use of the spin bit. The 
+pseudo provided in the appendix is based on input provided by David Schinazi.
 
 # Acknowledgments
 
+Thanks to Martin Thomas and Martin Duke for contributing by reviewing and
```
martinthomson, line 915: Who is this person?

draft-ietf-quic-manageability.md
```diff
@@ -863,11 +907,167 @@ security-relevant.
...
+\{0x51, 0x30, 0x35, 0x30\}. Otherwise, we use its hexadecimal representation:
+for example, 0xff00001d refers to {0xff, 0x00, 0x00, 0x1d}.
+
+QUIC versions that start with 'Q' or 'T' followed by three digits are 
+Google QUIC versions. Versions up to and including 43 are documented by
+<https://docs.google.com/document/d/1WJvyZflAO2pq77yOLbp9NsGjC1CHetAXV8I0fQe-B_U/preview>.
+Versions Q046, Q050, T050, and T051 are not fully documented, but this appendix
+should contain enough information to allow parsing Client Hellos for those versions.
+
+This appendis uses the following conventions:
```
martinthomson, line 943:
```suggestion
This appendix uses the following conventions:
```

martinthomson (commented):
I don't think that you should include the entirety of David's code.  There are ways in which that will contribute to ossification.  Indeed, I don't think that it is helpful to list draft versions here either.

draft-ietf-quic-manageability.md
```diff
@@ -587,6 +589,48 @@ Work is currently underway in the TLS working group to encrypt the SNI in TLS
 would make SNI-based application identification impossible through passive
 measurement.
 
+### Extracting Server Name Indication (SNI) Information
+
+If the SNI is not encrypted it can be derived form the QUIC Initial packet
```
britram, line 594:
```suggestion
If the SNI is not encrypted it can be derived from the QUIC Initial packet 
```

draft-ietf-quic-manageability.md
```diff
@@ -587,6 +589,48 @@ Work is currently underway in the TLS working group to encrypt the SNI in TLS
 would make SNI-based application identification impossible through passive
 measurement.
 
+### Extracting Server Name Indication (SNI) Information
+
+If the SNI is not encrypted it can be derived form the QUIC Initial packet 
+by calculating the Initial Secret to decrypt the packet payload and parse the 
+QUIC CRYPTO Frame with the TLS ClientHello.
```
britram, line 596:
```suggestion
QUIC CRYPTO Frame containing the TLS ClientHello.
```

draft-ietf-quic-manageability.md
```diff
@@ -863,11 +907,167 @@ security-relevant.
...
+providing text proposals.
+
 This work is partially supported by the European Commission under Horizon 2020
 grant agreement no. 688421 Measurement and Architecture for a Middleboxed
 Internet (MAMI), and by the Swiss State Secretariat for Education, Research, and
 Innovation under contract no. 15.0268. This support does not imply endorsement.
+
+# Appendix
+
+## Handling of Google QUIC  {#sec-google-version}
```
britram, line 925: I agree with @martinthomson here: beyond this section (mentioning the version numbers) and its first subsection (explaining how to differentiate Google and IETF QUIC), I'd scrub references to Google QUIC from the rest of the pseudocode, and split the appendix in two (first appendix: recognizing google quic; second appendix: initial salt (IETF only) and crypto frame extraction (IETF only)

draft-ietf-quic-manageability.md
```diff
@@ -863,11 +907,167 @@ security-relevant.
...
 Internet (MAMI), and by the Swiss State Secretariat for Education, Research, and
 Innovation under contract no. 15.0268. This support does not imply endorsement.
+
+# Appendix
+
+## Handling of Google QUIC  {#sec-google-version} 
+
+This sections contains algorithms that allows parsing versions from both
+Google QUIC and IETF QUIC. These mechanisms will become
+irrelevant when IETF QUIC is fully deployed and Google QUIC is deprecated.
```
britram, line 929: Add something here noting that the rest of this document says nothing about Google QUIC.

draft-ietf-quic-manageability.md
britram: ("with" sounds a little like one uses a ClientHello to parse the crypto frame, which would be a neat trick. :) )

draft-ietf-quic-manageability.md
```diff
@@ -587,6 +589,48 @@ Work is currently underway in the TLS working group to encrypt the SNI in TLS
...
+The source connection ID length is specified in the byte after the destination
+connection ID. And the token length, which follows the source connection ID,
+is a variable length integer as specified in section 16 of {{QUIC-TRANSPORT}}.
+
+Finally after decryption, the Initial Client packet can be parsed to detect the
+CRYPTO frame that contains the TLS Client Hello, which then can be respectively
+parsed similar as for all other TLS connections. The Initial client packet may
+contain other frames, so the first byte of each frame need to be checked to identify
+the frame type and the skip over the frame. Note that the length of the frames is 
+depended on the frame type. Usually for QUIC version 1, the packet is expected
```
britram, line 630:
```suggestion
dependent on the frame type. Usually for QUIC version 1, the packet is expected 
```

draft-ietf-quic-manageability.md
```diff
@@ -587,6 +589,48 @@ Work is currently underway in the TLS working group to encrypt the SNI in TLS
...
+is a variable length integer as specified in section 16 of {{QUIC-TRANSPORT}}.
+
+Finally after decryption, the Initial Client packet can be parsed to detect the
+CRYPTO frame that contains the TLS Client Hello, which then can be respectively
+parsed similar as for all other TLS connections. The Initial client packet may
+contain other frames, so the first byte of each frame need to be checked to identify
+the frame type and the skip over the frame. Note that the length of the frames is 
+depended on the frame type. Usually for QUIC version 1, the packet is expected 
+to only carry the CRYPTO frame and optionally padding frames. However, padding
+which is one byte of zeros, may occur after or also before the CRYPO frame.
```
britram, line 632:
```suggestion
which is one byte of zeros, may also occur before or after the CRYPTO frame.
```

having trouble parsing "padding which is one byte of zeroes" -- is the only allowable padding here exactly one byte long? haven't dug into this in a while.

britram (changes requested):
thanks for this! one major comment agreeing with @martinthomson: the appendix should drop some details of parsing Google QUIC, especially crypto frame extraction. otherwise a few nits.

draft-ietf-quic-manageability.md
mirjak: sorry, sorry, sorry

draft-ietf-quic-manageability.md
mirjak: Okay, I removed the salt section entirely (because I think that value should be in the tls draft and only there) and cut down the crypto frame part to IETF QUIC only.

draft-ietf-quic-manageability.md
```diff
@@ -863,11 +910,101 @@ security-relevant.
...
+This appendix uses the following conventions:
+array\[i\] – one byte at index i of array
+array\[i:j\] – subset of array starting with index i (inclusive) up to j-1
+(inclusive)
+array\[i:\] – subset of array starting with index i (inclusive) up to the
+end of the array
+
+## Distinguishing IETF QUIC and Google QUIC Versions {#sec-google-version}
+
+This sections contains algorithms that allows parsing versions from both
```
britram, line 937:
```suggestion
This section contains algorithms that allows parsing versions from both
```

draft-ietf-quic-manageability.md
```diff
@@ -863,11 +910,104 @@ security-relevant.
...
+end of the array
+
+## Distinguishing IETF QUIC and Google QUIC Versions {#sec-google-version}
+
+This section contains algorithms that allows parsing versions from both
+Google QUIC and IETF QUIC. These mechanisms will become
+irrelevant when IETF QUIC is fully deployed and Google QUIC is deprecated.
+
+Note that other than this appendix, nothing in this document applies to
+Google QUIC.
```
mirjak, line 942:
```suggestion
Google QUIC. And the purpose of this appendix is merely to distinguish IETF QUIC 
from any versions of Google QUIC.
```

britram:
fixes #75

britram (approved):

PR #126, opened 2020-10-27
Further text on idle timeout/keep alive frequency

mirjak:


draft-ietf-quic-applicability.md
```diff
@@ -240,9 +240,15 @@ keep-alives, to prevent the connection and any on-path state from timing out.
...
-much unproductive traffic and power usage for some situations.
+is responsible for keeping the application alive.  While {{Hatonen10}} suggests
+that 30 seconds might be a suitable value for the public Internet when a NAT
+is on path, larger values are preferable if the deployment can consistently
+survive NAT rebinds or is known to be in a controlled environments like e.g.
+data centres in order to lower network and computational load. Sending PING
+frames more frequently than every 30 seconds over long idle periods may result
+in a too much unproductive traffic and power usage for some situations. Further
+short time-outs can make it harder to handle short network interrupts, like VM
+mitragtion or coverage loss during mobilty.
```
ianswett, line 251: Wording suggestion, but I'm not sure what 'mobility' is in this context.  Is that moving connection migration(ie: Cell to Wifi) or something else?

```suggestion
data centres in order to lower network and computational load. Sending PING
frames more frequently than every 30 seconds over long idle periods may result
in a too much unproductive traffic and power usage for some situations. Additionally,
time-outs shorter than 30 seconds can make it harder to handle short network
interruptions, such as VM migration or coverage loss during mobilty.
```

draft-ietf-quic-applicability.md
```diff
@@ -240,9 +240,15 @@ keep-alives, to prevent the connection and any on-path state from timing out.
...
-is responsible for keeping the application alive. Note that sending PING frames
-more frequently than every 30 seconds over long idle periods may result in a too
-much unproductive traffic and power usage for some situations.
+is responsible for keeping the application alive.  While {{Hatonen10}} suggests
+that 30 seconds might be a suitable value for the public Internet when a NAT
+is on path, larger values are preferable if the deployment can consistently
+survive NAT rebinds or is known to be in a controlled environments like e.g.
+data centres in order to lower network and computational load. Sending PING
+frames more frequently than every 30 seconds over long idle periods may result
+in a too much unproductive traffic and power usage for some situations. Further
```
nibanks, line 249: I think `a` is unnecessary here.
```suggestion
in too much unproductive traffic and power usage for some situations. Further
```
Though `too much unproductive` might also be better replaced with `unacceptable` as well.

draft-ietf-quic-applicability.md
mirjak: No necessarily mitigation (but could be as well) but in taps we recently discussed this example where you walk into an elevator and loose connectivity for a few seconds but then are connected immediately again to the same network when you step out. Maybe that not worth mentioning explicitly?

draft-ietf-quic-applicability.md
mirjak: Yes, I was looking for better wording here as well!

Issue #115, opened 2020-07-31
connection migration: add server preferred address


martinduke:
We should probably make it clear what the application can do here (SPA for servers; maybe intentional migration for clients?) I'm not sure if migration is transparent to client applications or not.

mirjak:
The client application might have some knowledge if it's behind a NAT or not and as such might want to indicate to use Conn ID or not. We can make that more explicit in the text.

mirjak:
I did some small edits on the text in the connection migration section but @martinduke maybe you had something more in mind and want to propose more text?

mirjak:
This whole section should then also be moved to after the Conn ID section.

martinduke:
I like the move of text to the manageability draft. If it were me, I'd move "Server-Generated Connection ID" as well, as I don't think apps need to worry about it, but it's not a big deal.

I think adding text about server preferred address is probably more important for this draft.

mirjak:
Added some more text in #127

britram:
@martinduke we think the current state of the draft addresses this issue; could you check and close this issue if so?

PR #127, opened 2020-10-27
More text on SPA

mirjak:


draft-ietf-quic-applicability.md
```diff
@@ -466,6 +466,9 @@ based on the Connection ID (see also {{connid}}) in the QUIC header,
 if present. This supports cases where address information changes, such as
 NAT rebinding, intentional change of the local interface, or based on an
 indication in the handshake of the server for a preferred address to be used.
+As such if the client knows or is likely to sit behind a NAT, use of the 
+Connection ID is strongly recommended. Further use of the Connection ID is
+generally recommendated when mitragtion is supported.
```
martinthomson, line 471:
```suggestion
generally recommendated when migration is supported.
```

draft-ietf-quic-applicability.md
```diff
@@ -466,6 +466,9 @@ based on the Connection ID (see also {{connid}}) in the QUIC header,
 if present. This supports cases where address information changes, such as
 NAT rebinding, intentional change of the local interface, or based on an
 indication in the handshake of the server for a preferred address to be used.
+As such if the client knows or is likely to sit behind a NAT, use of the 
+Connection ID is strongly recommended. Further use of the Connection ID is
```
martinthomson, line 470:
```suggestion
Connection ID is strongly recommended. A non-empty the Connection ID is
```

draft-ietf-quic-applicability.md
martinthomson: This is pretty woolly.  I think that you want to say why: identifying the connection that packets are destined for is more challenging if they don't include a connection ID and the source address might have changed.  An empty connection ID is only appropriate if the local address is unique, which opens up some negative consequences in terms of providing obvious linkability (though in practice this only works with limited numbers of connections on the target IP+port, so linkability is pretty much a given at that point.)

draft-ietf-quic-applicability.md
```diff
@@ -466,6 +466,9 @@ based on the Connection ID (see also {{connid}}) in the QUIC header,
 if present. This supports cases where address information changes, such as
 NAT rebinding, intentional change of the local interface, or based on an
 indication in the handshake of the server for a preferred address to be used.
+As such if the client knows or is likely to sit behind a NAT, use of the 
+Connection ID is strongly recommended. A non-empty Connection ID is
+generally recommended when migration is supported.
```
mirjak, line 471: @martinthomson said:
`This is pretty woolly. I think that you want to say why: identifying the connection that packets are destined for is more challenging if they don't include a connection ID and the source address might have changed. An empty connection ID is only appropriate if the local address is unique, which opens up some negative consequences in terms of providing obvious linkability (though in practice this only works with limited numbers of connections on the target IP+port, so linkability is pretty much a given at that point.)`

I mainly wanted to give a recommendation that if you have no idea if you want/need a connection ID, you should probably have it (assuming that people agree that the right recommendation to give). I can add more explanation and also say something about linkability but I wanted to keep it simple... happy to take suggestions

draft-ietf-quic-applicability.md
mirjak: I mainly wanted to give a recommendation that if you have no idea if you want/need a connection ID, you should probably have it (assuming that people agree that the right recommendation to give). I can add more explanation and also say something about linkability but I wanted to keep it simple... happy to take suggestions

draft-ietf-quic-applicability.md
```diff
@@ -478,6 +481,13 @@ As long as the new path is not validated only probing packets can be sent.
 However, the probing packets can be used measure path characteristics as
 input for the switching decision or the congestion controller on the new path.
 
+Only the client can actively migrate, however, servers can indicate during the
+handshake that they prefer to transfer the connection to a different address
+after the handshake, e.g. to move from an address that is shared by multiple
+servers to an unicast address. The server can provide an IPv4 and an IPv6
+address as transport paramenter during the TLS and the client can select
+between the two if both are provided. See also Section 9.6 of {{!QUIC}}.
```
ekinnear, line 489: Tiny nit
```suggestion
address as transport paramenter during the TLS handshake and the client can 
select between the two if both are provided. See also Section 9.6 of {{!QUIC}}.
```

draft-ietf-quic-applicability.md
```diff
@@ -466,6 +466,9 @@ based on the Connection ID (see also {{connid}}) in the QUIC header,
 if present. This supports cases where address information changes, such as
 NAT rebinding, intentional change of the local interface, or based on an
 indication in the handshake of the server for a preferred address to be used.
+As such if the client knows or is likely to sit behind a NAT, use of the
+Connection ID is strongly recommended. A non-empty Connection ID is
```
martinthomson, line 470:
```suggestion
connection ID is strongly recommended. A non-empty connection ID is
```
I know that this likely offends some proper noun sensibilities, but we use lowercase for connection ID.

draft-ietf-quic-applicability.md
martinthomson:
```suggestion
address as transport parameter during the TLS handshake and the client can 
select between the two if both are provided. See also Section 9.6 of {{!QUIC}}.
```
Without the typo.

draft-ietf-quic-applicability.md
```diff
@@ -478,6 +481,13 @@ As long as the new path is not validated only probing packets can be sent.
 However, the probing packets can be used measure path characteristics as
 input for the switching decision or the congestion controller on the new path.
 
+Only the client can actively migrate, however, servers can indicate during the
+handshake that they prefer to transfer the connection to a different address
+after the handshake, e.g. to move from an address that is shared by multiple
```
martinthomson, line 486:
```suggestion
after the handshake, e.g., to move from an address that is shared by multiple
```

draft-ietf-quic-applicability.md
```diff
@@ -478,6 +481,13 @@ As long as the new path is not validated only probing packets can be sent.
...
+servers to an unicast address. The server can provide an IPv4 and an IPv6
```
martinthomson, line 487:
```suggestion
after the handshake, e.g., to move from an anycast address to a unicast
address. The server can provide an IPv4 and an IPv6
```

Though it might be better to instead say.

```suggestion
after the handshake, e.g. to move from an address that is shared by multiple
servers to an address that is unique to the server instance. The server can
provide an IPv4 and an IPv6
```

...as that allows for the possibility of both anycast -> unicast transitions and load balanced to non-load balanced transitions both.

draft-ietf-quic-applicability.md
```diff
@@ -466,6 +466,9 @@ based on the Connection ID (see also {{connid}}) in the QUIC header,
 if present. This supports cases where address information changes, such as
 NAT rebinding, intentional change of the local interface, or based on an
 indication in the handshake of the server for a preferred address to be used.
+As such if the client knows or is likely to sit behind a NAT, use of the
+connection ID is strongly recommended. A non-empty connection ID is
+generally recommended when migration is supported.
```
ianswett, line 471:
```suggestion
connection ID by the server is strongly recommended. A non-empty server
connection ID is also strongly recommended when migration is supported.
```

draft-ietf-quic-applicability.md
```diff
@@ -466,6 +466,9 @@ based on the Connection ID (see also {{connid}}) in the QUIC header,
 if present. This supports cases where address information changes, such as
 NAT rebinding, intentional change of the local interface, or based on an
 indication in the handshake of the server for a preferred address to be used.
+As such if the client knows or is likely to sit behind a NAT, use of the
```
ianswett, line 469: Even if the client knows, it can't communicate that to the server in v1.

```suggestion
As such if the client is known or likely to sit behind a NAT, use of the
```

draft-ietf-quic-applicability.md
```diff
@@ -478,6 +481,14 @@ As long as the new path is not validated only probing packets can be sent.
 However, the probing packets can be used measure path characteristics as
 input for the switching decision or the congestion controller on the new path.
 
+Only the client can actively migrate, however, servers can indicate during the
```
ianswett, line 484:
```suggestion
Only the client can actively migrate. However, servers can indicate during the
```

draft-ietf-quic-applicability.md
```diff
@@ -478,6 +481,14 @@ As long as the new path is not validated only probing packets can be sent.
...
+handshake that they prefer to transfer the connection to a different address
+after the handshake, e.g. to move from an address that is shared by multiple
+servers to an address that is unique to the server instance. The server can
+provide an IPv4 and an IPv6 address as transport parameter during the TLS
```
ianswett, line 488:
```suggestion
provide an IPv4 and an IPv6 address in a transport parameter during the TLS
```

britram:
note: fixes #115 when merged.

draft-ietf-quic-applicability.md
britram: We're going to go ahead and merge this for the 109 document deadline submission; please open another issue / provide a PR / haunt us otherwise if this needs more discussion.

PR #132, opened 2020-11-18
Remove normative language from the drafts

### LINKED ISSUE #85: Double check use of normative language
Not clear if these documents should use normative language at all. The applicability statement also sometimes uses normative language to recommend the implementation of certain interfaces; I don't think this is the right document for that but as we also don't have another document, we should decide if we want to do that or not!

britram:
fixes #85


britram:
note: don't merge yet, still need to make the pass on applicability but working on the CI issue at the moment

PR #134, opened 2020-11-18
Remove QOS support and ECMP

### LINKED ISSUE #131: QoS support and ECMP section
This section is labeled with \[EDITOR'S NOTE: this is a bit speculative; keep?]. Should we keep it?

britram:
proposed to fix #131 

(if we decide that's what we want to do)

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -877,30 +877,6 @@ and ACK manipulation is impossible. Specifically, heuristics attempting to
 distinguish ACK-only packets from payload-carrying packets based on packet size
 are likely to fail, and are emphatically NOT RECOMMENDED.
 
-## QoS support and ECMP
-
-\[EDITOR'S NOTE: this is a bit speculative; keep?]
-
-QUIC does not provide any additional information on requirements on Quality of
-Service (QoS) provided from the network. QUIC assumes that all packets with the
-same 5-tuple {dest addr, source addr, protocol, dest port, source port} will
-receive similar network treatment.  That means all stream that are multiplexed
-over the same QUIC connection require the same network treatment and are handled
-by the same congestion controller. If differential network treatment is desired,
-multiple QUIC connections to the same server might be used, given that
-establishing a new connection using 0-RTT support is cheap and fast.
-
-QoS mechanisms in the network MAY also use the connection ID for service
-differentiation, as a change of connection ID is bound to a change of address
-which anyway is likely to lead to a re-route on a different path with different
-network characteristics.
-
-Given that QUIC is more tolerant of packet re-ordering than TCP (see
-{{packetnumber}}), Equal-cost multi-path routing (ECMP) does not necessarily
-need to be flow based.  However, 5-tuple (plus eventually connection ID if
-present) matching is still beneficial for QoS given all packets are handled by
-the same congestion controller.
-
 # IANA Considerations
 
 This document has no actions for IANA.


draft-ietf-quic-manageability.md
```diff
@@ -877,30 +877,6 @@ and ACK manipulation is impossible. Specifically, heuristics attempting to
...
-receive similar network treatment.  That means all stream that are multiplexed
-over the same QUIC connection require the same network treatment and are handled
-by the same congestion controller. If differential network treatment is desired,
-multiple QUIC connections to the same server might be used, given that
-establishing a new connection using 0-RTT support is cheap and fast.
-
-QoS mechanisms in the network MAY also use the connection ID for service
-differentiation, as a change of connection ID is bound to a change of address
-which anyway is likely to lead to a re-route on a different path with different
-network characteristics.
```
martinthomson, line 896: This is likely bad advice anyway.  If we are going to signal intent to the network, it would be done explicitly (DSCP anyone?).  Inference is what gets us into messes.

martinthomson (approved):

draft-ietf-quic-manageability.md
```diff
@@ -877,30 +877,6 @@ and ACK manipulation is impossible. Specifically, heuristics attempting to
...
-\[EDITOR'S NOTE: this is a bit speculative; keep?]
-
-QUIC does not provide any additional information on requirements on Quality of
-Service (QoS) provided from the network. QUIC assumes that all packets with the
-same 5-tuple {dest addr, source addr, protocol, dest port, source port} will
-receive similar network treatment.  That means all stream that are multiplexed
-over the same QUIC connection require the same network treatment and are handled
-by the same congestion controller. If differential network treatment is desired,
-multiple QUIC connections to the same server might be used, given that
-establishing a new connection using 0-RTT support is cheap and fast.
```
mirjak, line 891: I think we should keep this first part here. Especially the part saying "If differential network treatment is desired,
 multiple QUIC connections to the same server might be used" because the idea that QUIC packets of the same connection that carry different stream data could have different DCSPs came up more than once that that really is a bad idea for CC.

draft-ietf-quic-manageability.md
mirjak: This part definitely needs rewriting or removing.

draft-ietf-quic-manageability.md
```diff
@@ -877,30 +877,6 @@ and ACK manipulation is impossible. Specifically, heuristics attempting to
...
-QoS mechanisms in the network MAY also use the connection ID for service
-differentiation, as a change of connection ID is bound to a change of address
-which anyway is likely to lead to a re-route on a different path with different
-network characteristics.
-
-Given that QUIC is more tolerant of packet re-ordering than TCP (see
-{{packetnumber}}), Equal-cost multi-path routing (ECMP) does not necessarily
-need to be flow based.  However, 5-tuple (plus eventually connection ID if
-present) matching is still beneficial for QoS given all packets are handled by
-the same congestion controller.
```
mirjak, line 902: This is probably also fine (minus the ConnID part) and probably clearing that not QUIC connections might be equally tolerant to re-ordering. It's just more likely than for TCP that the implementation does that.

draft-ietf-quic-manageability.md
martinthomson: That is probably a deep enough conversation that I don't think it is worth including here.  Multiple connections might not give you what you want either.  And different congestion controllers might tolerate variations in QoS better.  The problem with all of this text is that it is creating the impression that underlying semantics can be inferred.  As a general theme, that is not good and so saying anything more than what is strictly factual is likely to encourage ossification of various sorts.

draft-ietf-quic-manageability.md
martinthomson: This is equally bad.  It says that you can ECMP as you please and QUIC will struggle on.  Now, this is probably out of endpoint control anyway, but that doesn't need to be endorsed.  The assumption should be that flows get uniform treatment unless they explicitly request otherwise.

draft-ietf-quic-manageability.md
mirjak: So what I would like to say somewhere if that if an endpoint want different network treatment based on DSCP (ignoring that DSCP is not usable e2e) it should use different connections and set the same DSCP on all packets of the same connection, instead of getting the idea of setting different DSCP on different packets of the same connection.

mirjak: Also I think the guidance for ECMP is still to schedule based on 5-tuple (and not on a per-packet base). I think that is still true for QUIC and we should state that.

As such if an endpoint wants to somehow make use of ECMP (e.g. in a datacenter setup) on need to consider to use different connections.

I think we can also say that it is generally recommended to use as few QUIC connections as possible to the same server (for CC reasons).

draft-ietf-quic-manageability.md
mirjak: Yes that's what the text was actually trying to say. As I said I think we need to re-phrase though.

britram:
Rolled over to #140, @mirjak to rewrite

PR #135, opened 2020-11-19
Edits, first batch

martinthomson:
These are pretty big.  If you would prefer that I create a PR for each individual commit, that can be done.

It turns out that there were a lot of things I found needed work.  For instance, this completely redoes the 0-RTT advice.

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -223,17 +223,21 @@ requires a minimum value of 15 seconds, but recommends larger values, or
 omitting keepalive entirely.
 
 By using a Connection ID, QUIC is designed to be robust to NAT address
-rebinding after a timeout. However, some QUIC connections may not be robust to
-rebinding because the routing infrastructure (in particular, load balancers)
-uses the address/port four-tuple to direct traffic. Furthermore, middleboxes
-with functions other than address translation may still affect the path. In
-particular, firewalls will often not admit server traffic for which it has not
-kept state for corresponding packets from the client.
+rebinding after a timeout. However, this only helps if one endpoint maintains
+availability at the address its peer uses, and the peer is the one to send
+after the timeout occurs.
 
-A QUIC application has three strategies to deal with this issue by
-adjusting idle periods (noting that idle periods and the network idle
-timeout is distinct from the connection idle timeout, defined as the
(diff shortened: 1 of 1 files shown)


britram (approved):
Aside from the keep-alive tweak and the idempotence change, these are all editorial and LGTM, thanks!

The keepalive tweak is also good.

Having looked at https://github.com/quicwg/base-drafts/pull/4394, the text here is a good applicability-appropriate view on what's there now. assuming discussion there converges to mostly the same thing, I'd be happy to merge this.

(in the future, no need to make tiny PRs, but splitting into "editorial-rollup" and "substantive editorial changes" would be nice to separate discussion, since the former shouldn't need much of it)

martinthomson:
It did kinda snowball, for which I apologize.  I was trying to get something under your nose before I got swamped by meetings.  I am also on vacation next week.

I'm reading through and there is some really good stuff in the doc.  I have started to open issues where I don't think the fixes are so simple.

draft-ietf-quic-applicability.md
```diff
@@ -163,46 +166,45 @@ and challenges for applications using QUIC.
...
-## Here There Be Dragons
+Application protocols that use 0-RTT require a profile that describes the types
+of information that can be safely sent. For HTTP, this profile is described in
+{{?HTTP-REPLAY=RFC8470}}.
 
-Retransmission or (malicious) replay of data contained in 0-RTT resumption
-packets could cause the server side to receive two copies of the same data.
-This is further described in {{?HTTP-RETRY=I-D.nottingham-httpbis-retry}}.
-Data sent during 0-RTT resumption also cannot benefit from perfect forward
-secrecy (PFS).
```
mirjak, line 185: The new text does not mention PFS anymore. Do we need to retain that?

mirjak (approved):
Changes look all ready good. Thanks!

I just have left one minor comment/question to double-check before we merge.

draft-ietf-quic-applicability.md
martinthomson: Forward secrecy (or post-compromise security to be precise) can be provided for 0-RTT under some conditions, though they are quite unusual in practice.  Generally, we wouldn't assume PFS for 0-RTT, but you can't claim that it is impossible to achieve.

PR #150, opened 2020-12-10
Editorial Pass on quic-manageability

martinduke:
This has some updates and clarifications based on protocol evolution, which might be uncontroversial. There are also many typo fixes.

PTAL

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -73,11 +73,10 @@ of QUIC-aware middleboxes, e.g. for load balancing.
 # Introduction
 
 QUIC {{?QUIC-TRANSPORT=I-D.ietf-quic-transport}} is a new transport protocol
-focusing on support of semantics as needed for HTTP
-{{?QUIC-HTTP=I-D.ietf-quic-http}}.  QUIC is encapsulated in UDP and encrypted
-by default. QUIC integrates TLS {{?QUIC-TLS=I-D.ietf-quic-tls}} to encrypt all
-payload data and most control information. The version of HTTP that operates
-over QUIC is HTTP/3.
+encapsulated in UDP and encrypted by default. QUIC integrates TLS
+{{?QUIC-TLS=I-D.ietf-quic-tls}} to encrypt all payload data and most control
+information. The design focused on support of semantics for HTTP, which required
+changes to HTTP known as HTTP/3 {{?QUIC-HTTP=I-D.ietf-quic-http}}.
 
 Given that QUIC is an end-to-end transport protocol, all information in the
 protocol header, even that which can be inspected, is not meant to be
@@ -153,8 +152,7 @@ The following information is exposed in QUIC packet headers:
 - "fixed bit": the second most significant bit of the first octet most QUIC
   packets of the current version is currently set to 1, for demultiplexing
   with other UDP-encapsulated protocols. Endpoints that do not need this
-  demultiplexing function might assign arbitrary values. See
-  {{I-D.thomson-quic-bit-grease}} for more details.
+  demultiplexing function might negotiate the use of arbitrary values.
 
 - latency spin bit: the third most significant bit of first octet in the short
   packet header. The spin bit is set by endpoints such that  tracking edge
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -576,7 +576,7 @@ packet.
 
 Note that clients send Initial packets before servers do, servers send Handshake
 packets before clients do, and only clients send Initial packets with tokens.
-Therefore, identity as a client or server can generally be confirmed by an on-
+Therefore, the role as a client or server can generally be confirmed by an on-
 path observer. An attempted connection after Retry can be detected by
 correlating the token on the Retry with the token on the subsequent Initial
 packet and the Destination Connection ID of the new Initial packet.
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -782,7 +782,7 @@ timers or through least-recently-used eviction, depending on application
 requirements.
 
 {{?RFC4787}} recommends a 2 minute timeout interval for UDP. However,
-timers are often lower, in the range of 15 to 30 seconds. In contrast,
+timers can be lower, in the range of 15 to 30 seconds. In contrast,
 {{?RFC5382}} recommends a timeout of more than 2 hours for TCP, given that TCP
 is a connection-oriented protocol with well-defined closure semantics. For
 network devices that are QUIC-aware, it is recommended to also use
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -640,7 +640,7 @@ PADDING frames, which are each one byte of zeros, may also occur before or after
 the CRYPTO frame.
 
 Note that client Initial packets that do not contain the Client Hello message
-generally also do not always use the Destination Connection ID that was used to
+do not always use the Destination Connection ID that was used to
 generate the Initial keys. Therefore, attempts to decrypt these packets using
 the procedure above will fail.
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -73,34 +73,33 @@ of QUIC-aware middleboxes, e.g. for load balancing.
 # Introduction
 
 QUIC {{?QUIC-TRANSPORT=I-D.ietf-quic-transport}} is a new transport protocol
-currently under development in the IETF QUIC working group, focusing on
-support of semantics as needed for HTTP/2 {{?QUIC-HTTP=I-D.ietf-quic-http}}.
-Based on current deployment practices, QUIC is encapsulated in UDP and
-encrypted by default. The current version of QUIC integrates TLS
-{{?QUIC-TLS=I-D.ietf-quic-tls}} to encrypt all payload data and most control
-information.
+focusing on support of semantics as needed for HTTP
+{{?QUIC-HTTP=I-D.ietf-quic-http}}.  QUIC is encapsulated in UDP and encrypted
+by default. QUIC integrates TLS {{?QUIC-TLS=I-D.ietf-quic-tls}} to encrypt all
+payload data and most control information. The version of HTTP that operates
+over QUIC is HTTP/3.
 
 Given that QUIC is an end-to-end transport protocol, all information in the
 protocol header, even that which can be inspected, is not meant to be
 mutable by the network, and is therefore integrity-protected. While less
 information is visible to the network than for TCP, integrity protection can
-also simplify troubleshooting because none of the nodes on the network path
+also simplify troubleshooting, because none of the nodes on the network path
 can modify the transport layer information.
 
-This document provides guidance for network operation on the management of QUIC
+This document provides guidance for network operations that manage of QUIC
 traffic. This includes guidance on how to interpret and utilize information that
-is exposed by QUIC to the network as well as explaining requirement and
-assumptions that the QUIC protocol design takes toward the expected network
-treatment. It also discusses how common network management practices will be
-impacted by QUIC.
+is exposed by QUIC to the network, requirements and assumptions that the QUIC
+design with respect to network treatment, and a description of how common
+network management practices will be impacted by QUIC.
 
 Since QUIC's wire image {{?WIRE-IMAGE=RFC8546}} is integrity protected and not
 modifiable on path, in-network operations are not possible without terminating
 the QUIC connection, for instance using a back-to-back proxy. Proxy operations
 are not in scope for this document. QUIC proxies must be fully-fledged QUIC
 endpoints, implementing the transport as defined in {{QUIC-TRANSPORT}} and
-{{QUIC-TLS}} as well as proxy-relevant semantics for the application(s) running
-over QUIC (e.g. HTTP/3 as defined in {{QUIC-HTTP}}).
+{{QUIC-TLS}}, proxy-relevant semantics for the application(s) running over QUIC
+(e.g. HTTP/3 as defined in {{QUIC-HTTP}}), and have the TLS credentials to
+authenticate as the server and (in some cases) client.
 
 Network management is not a one-size-fits-all endeavour: practices considered
 necessary or even mandatory within enterprise networks with certain compliance
@@ -110,36 +109,33 @@ recommendations as to which practices should or should not be applied; for each
 practice, it describes what is and is not possible with the QUIC transport
 protocol as defined.
 
-QUIC is at the moment very much a moving target. This document refers the state
-of the QUIC working group drafts as well as to changes under discussion, via
-issues and pull requests in GitHub current as of the time of writing.
-
 # Features of the QUIC Wire Image {#sec-wire-image}
 
-In this section, we discusses those aspects of the QUIC transport protocol that
+In this section, we discuss those aspects of the QUIC transport protocol that
 have an impact on the design and operation of devices that forward QUIC packets.
 Here, we are concerned primarily with the unencrypted part of QUIC's wire image
 {{WIRE-IMAGE}}, which we define as the information available in the packet
 header in each QUIC packet, and the dynamics of that information. Since QUIC is
 a versioned protocol, the wire image of the header format can also change from
-version to version. However, at least the mechanism by which a receiver can
-determine which version is used and the meaning and location of fields used in
-the version negotiation process is invariant
-{{?QUIC-INVARIANTS=I-D.ietf-quic-invariants}}.
+version to version. However, the field that identifies the QUIC version in some
+packets, and the format of the Version Negotiation Packet, are both inspectable
+and invariant {{?QUIC-INVARIANTS=I-D.ietf-quic-invariants}}.
 
-This document describes only version 1 the QUIC protocol, whose wire image
-is fully defined in {{QUIC-TRANSPORT}} and {{?QUIC-TLS}}. Note that features
-of the wire image described herein and in those documents may change in future
-versions of the protocol, and cannot be used to identify QUIC as a protocol or
-to infer the behavior of future versions of QUIC. {{sec-google-version}}
-provides non-normative guidance on the identification of QUIC version 1
-packets compared to other deployed versions at the date if publication.
+This document describes version 1 of the QUIC protocol, whose wire image
+is fully defined in {{QUIC-TRANSPORT}} and {{?QUIC-TLS}}. Features of the wire
+ image described herein may change in future versions of the protocol, except
+when specified as an invariant {{QUIC-INVARIANTS}},
+and cannot be used to identify QUIC as a protocol or
+to infer the behavior of future versions of QUIC.
+
+{{sec-google-version}} provides non-normative guidance on the identification of
+QUIC version 1 packets compared to pre-standard versions.
 
 ## QUIC Packet Header Structure {#public-header}
 
 QUIC packets may have either a long header, or a short header. The first bit
-of the QUIC header us the Header Form bit, and indicates which type of header
-is present.
+of the QUIC header is the Header Form bit, and indicates which type of header
+is present. The purpose of this bit is invariant across QUIC versions.
 
 The long header exposes more information. It is used during connection
 establishment, including version negotiation, retry, and 0-RTT data. It
@@ -156,7 +152,9 @@ The following information is exposed in QUIC packet headers:
 
 - "fixed bit": the second most significant bit of the first octet most QUIC
   packets of the current version is currently set to 1, for demultiplexing
-  with other UDP-encapsulated protocols.
+  with other UDP-encapsulated protocols. Endpoints that do not need this
+  demultiplexing function might assign arbitrary values. See
+  {{I-D.thomson-quic-bit-grease}} for more details.
 
 - latency spin bit: the third most significant bit of first octet in the short
   packet header. The spin bit is set by endpoints such that  tracking edge
@@ -167,13 +165,15 @@ The following information is exposed in QUIC packet headers:
   Header Form and fixed bits. Header types correspond to stages of the
   handshake; see Section 17.2 of {{QUIC-TRANSPORT}} for details.
 
-- version number: the version number present in the long header, and identifies
+- version number: the version number is present in the long header, and identifies
   the version used for that packet. Note that during Version Negotiation (see
-  {{version}}, and Section 17.2.1 of {{QUIC-TRANSPORT}}, the version number
+  {{version}} and Section 17.2.1 of {{QUIC-TRANSPORT}}), the version number
   field has a special value (0x00000000) that identifies the packet as a Version
   Negotiation packet. QUIC versions that start with 0xff are IETF drafts. QUIC
   versions that start with 0x0000 are reserved for IETF consensus documents,
-  for example the QUIC version 1 is expected to use version 0x00000001.
+  for example the QUIC version 1 is expected to use version 0x00000001. Operators
+  should expect to observe packets with other version numbers as a result of
+  various internet experiments and future standards.
 
 - source and destination connection ID: short and long packet headers carry a
   destination connection ID, a variable-length field that can be used to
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -73,34 +73,33 @@ of QUIC-aware middleboxes, e.g. for load balancing.
...
-support of semantics as needed for HTTP/2 {{?QUIC-HTTP=I-D.ietf-quic-http}}.
-Based on current deployment practices, QUIC is encapsulated in UDP and
-encrypted by default. The current version of QUIC integrates TLS
-{{?QUIC-TLS=I-D.ietf-quic-tls}} to encrypt all payload data and most control
-information.
+focusing on support of semantics as needed for HTTP
+{{?QUIC-HTTP=I-D.ietf-quic-http}}.  QUIC is encapsulated in UDP and encrypted
+by default. QUIC integrates TLS {{?QUIC-TLS=I-D.ietf-quic-tls}} to encrypt all
+payload data and most control information. The version of HTTP that operates
+over QUIC is HTTP/3.
```
martinthomson, line 80: The two mentions of h3 here are disjoint.  Suggest moving the "semantics as needed for HTTP(/3)" later.

draft-ietf-quic-manageability.md
```diff
@@ -110,36 +109,33 @@ recommendations as to which practices should or should not be applied; for each
...
+
+This document describes version 1 of the QUIC protocol, whose wire image
+is fully defined in {{QUIC-TRANSPORT}} and {{?QUIC-TLS}}. Features of the wire
+ image described herein may change in future versions of the protocol, except
+when specified as an invariant {{QUIC-INVARIANTS}},
+and cannot be used to identify QUIC as a protocol or
+to infer the behavior of future versions of QUIC.
+
+{{sec-google-version}} provides non-normative guidance on the identification of
+QUIC version 1 packets compared to pre-standard versions.
```
martinthomson, line 132:
```suggestion
QUIC version 1 packets compared to some pre-standard versions.
```

No mention there of Facebook's versions, for instance.

draft-ietf-quic-manageability.md
```diff
@@ -156,7 +152,9 @@ The following information is exposed in QUIC packet headers:
 
 - "fixed bit": the second most significant bit of the first octet most QUIC
   packets of the current version is currently set to 1, for demultiplexing
-  with other UDP-encapsulated protocols.
+  with other UDP-encapsulated protocols. Endpoints that do not need this
+  demultiplexing function might assign arbitrary values. See
+  {{I-D.thomson-quic-bit-grease}} for more details.
```
martinthomson, line 157: Not sure if this is completely wise before it is adopted.  Of course, the statement will remain true nonetheless, even if that is never standardized.

draft-ietf-quic-manageability.md
```diff
@@ -167,13 +165,15 @@ The following information is exposed in QUIC packet headers:
...
-- version number: the version number present in the long header, and identifies
+- version number: the version number is present in the long header, and identifies
   the version used for that packet. Note that during Version Negotiation (see
-  {{version}}, and Section 17.2.1 of {{QUIC-TRANSPORT}}, the version number
+  {{version}} and Section 17.2.1 of {{QUIC-TRANSPORT}}), the version number
   field has a special value (0x00000000) that identifies the packet as a Version
   Negotiation packet. QUIC versions that start with 0xff are IETF drafts. QUIC
   versions that start with 0x0000 are reserved for IETF consensus documents,
-  for example the QUIC version 1 is expected to use version 0x00000001.
+  for example the QUIC version 1 is expected to use version 0x00000001. Operators
```
martinthomson, line 174:
```suggestion
  For example the QUIC version 1 uses version 0x00000001. Operators
```

If you are removing "QUIC is a moving target", then this will be the case.

Unless it isn't, but that will be a minor catastrophe that we can work through if it happens.

draft-ietf-quic-manageability.md
```diff
@@ -230,7 +230,7 @@ TCP-based services, especially when application layer information is
 encrypted, there is no guarantee that a specific application will use the
 registered port, or the used port is carrying traffic belonging to the
 respective registered service. For example, {{QUIC-TRANSPORT}} specifies
-the use of Alt-Svc for discovery of QUIC/HTTP services on other ports.
+the use of Alt-Svc for discovery of QUIC/HTTP3 services on other ports.
```
martinthomson, line 233:
```suggestion
the use of Alt-Svc for discovery of HTTP/3 services on other ports.
```

draft-ietf-quic-manageability.md
```diff
@@ -230,7 +230,7 @@ TCP-based services, especially when application layer information is
 encrypted, there is no guarantee that a specific application will use the
 registered port, or the used port is carrying traffic belonging to the
 respective registered service. For example, {{QUIC-TRANSPORT}} specifies
```
martinthomson, line 232: Alt-Svc use is defined in QUIC-HTTP:

```suggestion
respective registered service. For example, {{QUIC-HTTP}} specifies
```

draft-ietf-quic-manageability.md
```diff
@@ -286,10 +286,17 @@ datagram as shown in {{fig-server-hello}} typically containing three packets:
 an Initial packet with the Server Hello, a Handshake packet with the rest of
 the server's side of the TLS handshake, and initial 1-RTT data, if present.
 
+The Client Hello datagram, Server Hello datagram, and Initial Completion datagram,
+which all contain a QUIC Initial Packet, must be at least 1200 octets long. This
```
martinthomson, line 290: It is possible that the client will not include an Initial packet in its second flight.  Perhaps more importantly, if the client sends a Handshake in the same datagram, it is not required to pad that datagram.

draft-ietf-quic-manageability.md
```diff
@@ -476,26 +482,26 @@ therefore not part of the wire image that is visible to on-path observers.
 
 ## Version Negotiation and Greasing {#version}
 
-Version negotiation is not protected, given the used protection mechanism can
-change with the version.  However, the choices provided in the list of version
-in the Version Negotiation packet will be validated as soon as the cryptographic
-context has been established. Therefore any manipulation of this list will be
-detected and will cause the endpoints to terminate the connection.
+Version negotiation packets are not intrinsically protected, but QUIC versions
```
martinthomson, line 485:
```suggestion
Version Negotiation packets are not intrinsically protected, but QUIC versions
```

draft-ietf-quic-manageability.md
```diff
@@ -476,26 +482,26 @@ therefore not part of the wire image that is visible to on-path observers.
...
+can use later encrypted messages to verify that they were authentic.
+Therefore any manipulation of this list will be detected and will cause the
+endpoints to terminate the connection.
```
martinthomson, line 488: Not sure if this is the reaction is correct.  "The connection" isn't a notion that makes sense when VN is involved because there are multiple connection attempts involved: one that elicited the VN, and one that used the VN.  Neither will result in a connection being established if there was tampering.

Maybe "might terminate the connection attempt".

draft-ietf-quic-manageability.md
```diff
@@ -557,19 +564,20 @@ future versions of the protocol.
 
 ## Connection confirmation {#sec-confirm}
 
-Connection establishment uses Initial, Handshake, and Retry packets containing a
+Connection establishment uses Initial and Handshake packets containing a
```
martinthomson, line 567: It really does include Retry.  Though those don't include TLS handshake messages.

draft-ietf-quic-manageability.md
```diff
@@ -557,19 +564,20 @@ future versions of the protocol.
...
 TLS handshake. Connection establishment can therefore be detected using
-heuristics similar to those used to detect TLS over TCP. A client using 0-RTT
-connection may also send data packets in 0-RTT Protected packets directly after
-the Initial packet containing the TLS Client Hello. Since these packets may be
-reordered in the network, note that 0-RTT Protected data packets may be seen
-before the Initial packet.
+heuristics similar to those used to detect TLS over TCP. A client initiating
```
martinthomson, line 569:
```suggestion
heuristics similar to those used to detect TLS over TCP. A client initiating a
```

draft-ietf-quic-manageability.md
```diff
@@ -557,19 +564,20 @@ future versions of the protocol.
...
+0-RTT connection may also send data packets in 0-RTT Protected packets directly
+after the Initial packet containing the TLS Client Hello. Since these packets
+may be reordered in the network, note that 0-RTT Protected data packets may be
```
martinthomson, line 572:
```suggestion
may be reordered in the network, 0-RTT Protected data packets could be
```

draft-ietf-quic-manageability.md
```diff
@@ -612,24 +620,29 @@ version specific initial salt, as described in {{QUIC-TLS}}. The length of
...
-Finally after decryption, the Initial Client packet can be parsed to detect the
-CRYPTO frame that contains the TLS Client Hello, which then can be respectively
-parsed similar as for all other TLS connections. The Initial client packet may
-contain other frames, so the first byte of each frame need to be checked to
-identify the frame type and the skip over the frame. Note that the length of the
-frames is dependent on the frame type. Usually for QUIC version 1, the packet is
+{{QUIC-TRANSPORT}}, but is obfuscated as described in {{QUIC-TLS}}. The source
+connection ID length is specified in the byte after the destination connection
+ID. And the token length, which follows the source connection ID, is a variable
+length integer as specified in section 16 of {{QUIC-TRANSPORT}}.
```
martinthomson, line 630:
```suggestion
length integer as specified in Section 16 of {{QUIC-TRANSPORT}}.
```

draft-ietf-quic-manageability.md
```diff
@@ -612,24 +620,29 @@ version specific initial salt, as described in {{QUIC-TLS}}. The length of
...
+contain other frames, so the first bytes of each frame need to be checked to
+identify the frame type, and if needed skip over it. Note that the length of the
+frames is dependent on the frame type. In QUIC version 1, the packet is
 expected to only carry the CRYPTO frame and optionally padding frames. However,
-padding which is one byte of zeros, may also occur before or after the CRYPTO
-frame.
+PADDING frames, which are each one byte of zeros, may also occur before or after
+the CRYPTO frame.
+
+Note that client Initial packets that do not contain the Client Hello message
```
martinthomson, line 642: I read this at first and thought "wat?", but I think that you just have an extra negation here:

```suggestion
Note that client Initial packets that contain the Client Hello message
```

draft-ietf-quic-manageability.md
```diff
@@ -612,24 +620,29 @@ version specific initial salt, as described in {{QUIC-TLS}}. The length of
...
+generally also do not always use the Destination Connection ID that was used to
```
martinthomson, line 643:
```suggestion
do not always use the Destination Connection ID that was used to
```

draft-ietf-quic-manageability.md
```diff
@@ -767,12 +781,12 @@ signal ({{sec-teardown}}) means that this state must be purged either through
...
 requirements.
 
-{{?RFC4787}} recommends a 2 minute timeout interval for UDP, however,
-often timer are lower in the range of 15 to 30 second. In constrast
-{{?RFC5382}} recommends a timeout of more than 2 hours for TCP, given TCP is
-a connection-oriented protocol with well defined closure semantics. For network
-devices that are QUIC-aware, it is recommended to also use
-longer timeouts for QUIC traffic, as QUIC is connection-oriented and as such
+{{?RFC4787}} recommends a 2 minute timeout interval for UDP. However,
+timers are often lower, in the range of 15 to 30 seconds. In contrast,
```
martinthomson, line 785:
```suggestion
timers can be lower, in the range of 15 to 30 seconds. In contrast,
```

draft-ietf-quic-manageability.md
mirjak: Yes, I agree that we should not cite the draft here. However, we could say that the network cannot rely on this bit as the endpoint might decide to not support it (if not needed).

draft-ietf-quic-manageability.md
mirjak: I would move only the reference to the quic-http draft. Because the first mention if https is the protocol in general, while later it specifically takes about h3.

draft-ietf-quic-manageability.md
mirjak: I think "often" is correct here but fine...

mirjak (changes requested):
Thanks. These look good. I agree with MT's comments and suggestions.

draft-ietf-quic-manageability.md
martinduke: What I'm trying to refer to here is the Initial ACK. Observers should expect that a stateless DPI will like fail to decrypt ~50% of client Initial packets. I think this proposed change inverts the meaning?

draft-ietf-quic-manageability.md
martinduke: Did we make the Initial ACK formally optional, or are clients just omitting it as an optimization?

Sec 14.1 says "A client MUST expand the payload of all UDP datagrams carrying Initial packets to at least the smallest allowed maximum datagram size of 1200 bytes by adding PADDING frames to the Initial packet or by coalescing the Initial packet" without any qualification about Handshake.

My reading of the text is that it has to be 1200B, whether via PADDING or coalescing up to that point. I understand that there is little point to requiring that padding.

martinduke:
Alright, everything is resolved except two MT comments about the client Initial ACK.

- How do we phrase correctly that these cannot be decrypted by using the DCID?
- Do these datagrams need to be 1200B or not? (I contend that there is no real need, but that's not what the spec says).

draft-ietf-quic-manageability.md
martinthomson: ACK.  I just misread this.  The double-negation caught me out.  I think that you might want to try to work on this.  Maybe "client Initial packets other than those in the very first flight sent by the client"

It's got nothing to do with the ClientHello, it's whether the client has seen packets from the server yet.

draft-ietf-quic-manageability.md
martinthomson: Yes, if the client sends an Initial packet, it needs to be padded.  That's pretty clear.  My point is that it might not send it and so not cause that rule to apply.

Let's say that the client does send an Initial ACK before it discards keys.  That is lost.  It doesn't retransmit it because the keys are gone.  The observer won't see that; they only see the Handshake+ packet that is sent later.

And if the client does discard keys too soon, the protocol doesn't stop working and no rules have been broken.

I'm just asking that you be a little more careful and maybe say that a handshake will look like this *in most cases*.  There are too many ways this could be wrong and you used "must".

martinduke: OK, rewrote it, PTAL

draft-ietf-quic-manageability.md
```diff
@@ -286,10 +284,20 @@ datagram as shown in {{fig-server-hello}} typically containing three packets:
 an Initial packet with the Server Hello, a Handshake packet with the rest of
 the server's side of the TLS handshake, and initial 1-RTT data, if present.
 
+The Initial Completion datagram contains at least one Handshake packet and
+some also include an Initial packet.
+
+Datagrams that contain a QUIC Initial Packet (Client Hello, Server Hello, and
+some Initial Completion) must be at least 1200 octets long. This protects
+against amplification attacks and verifies the network path meets minimum
```
martinthomson, line 292:
```suggestion
against amplification attacks and verifies that the network path meets minimum
```

martinthomson (approved):

draft-ietf-quic-manageability.md
```diff
@@ -73,34 +73,32 @@ of QUIC-aware middleboxes, e.g. for load balancing.
...
 Since QUIC's wire image {{?WIRE-IMAGE=RFC8546}} is integrity protected and not
 modifiable on path, in-network operations are not possible without terminating
 the QUIC connection, for instance using a back-to-back proxy. Proxy operations
 are not in scope for this document. QUIC proxies must be fully-fledged QUIC
 endpoints, implementing the transport as defined in {{QUIC-TRANSPORT}} and
-{{QUIC-TLS}} as well as proxy-relevant semantics for the application(s) running
-over QUIC (e.g. HTTP/3 as defined in {{QUIC-HTTP}}).
+{{QUIC-TLS}}, proxy-relevant semantics for the application(s) running over QUIC
+(e.g. HTTP/3 as defined in {{QUIC-HTTP}}), and have the TLS credentials to
+authenticate as the server and (in some cases) client.
```
mirjak, line 101: Actually not sure here. The assumption wasn't necessary that the proxy can authenticate as the server (because then I guess it is the server) but probably rather that it would authenticate as an own entity offering a proxy service.

draft-ietf-quic-manageability.md
```diff
@@ -110,36 +108,33 @@ recommendations as to which practices should or should not be applied; for each
...
-to infer the behavior of future versions of QUIC. {{sec-google-version}}
-provides non-normative guidance on the identification of QUIC version 1
-packets compared to other deployed versions at the date if publication.
+version to version. However, the field that identifies the QUIC version in some
+packets, and the format of the Version Negotiation Packet, are both inspectable
+and invariant {{?QUIC-INVARIANTS=I-D.ietf-quic-invariants}}.
+
+This document describes version 1 of the QUIC protocol, whose wire image
+is fully defined in {{QUIC-TRANSPORT}} and {{?QUIC-TLS}}. Features of the wire
+ image described herein may change in future versions of the protocol, except
```
mirjak, line 125:
```suggestion
image described herein may change in future versions of the protocol, except
```
remove leading space...

draft-ietf-quic-manageability.md
```diff
@@ -156,7 +151,8 @@ The following information is exposed in QUIC packet headers:
 
 - "fixed bit": the second most significant bit of the first octet most QUIC
   packets of the current version is currently set to 1, for demultiplexing
-  with other UDP-encapsulated protocols.
+  with other UDP-encapsulated protocols. Endpoints that do not need this
+  demultiplexing function might negotiate the use of arbitrary values.
```
mirjak, line 155: Actually this new sentence does not seem right. The transport draft says:

```
   Fixed Bit:  The next bit (0x40) of byte 0 is set to 1.  Packets
      containing a zero value for this bit are not valid packets in this
      version and MUST be discarded.
```

draft-ietf-quic-manageability.md
```diff
@@ -167,13 +163,15 @@ The following information is exposed in QUIC packet headers:
...
 
-- version number: the version number present in the long header, and identifies
-  the version used for that packet. Note that during Version Negotiation (see
-  {{version}}, and Section 17.2.1 of {{QUIC-TRANSPORT}}, the version number
-  field has a special value (0x00000000) that identifies the packet as a Version
-  Negotiation packet. QUIC versions that start with 0xff are IETF drafts. QUIC
-  versions that start with 0x0000 are reserved for IETF consensus documents,
-  for example the QUIC version 1 is expected to use version 0x00000001.
+- version number: the version number is present in the long header, and
+  identifies the version used for that packet. Note that during Version
```
mirjak, line 167:
```suggestion
  identifies the version used for that packet. During Version
```
```suggestion
  identifies the version used for that packet. Note that during Version
```

draft-ietf-quic-manageability.md
```diff
@@ -167,13 +163,15 @@ The following information is exposed in QUIC packet headers:
...
+  Negotiation (see {{version}} and Section 17.2.1 of {{QUIC-TRANSPORT}}), the
+  version number field has a special value (0x00000000) that identifies the
+  packet as a Version Negotiation packet. QUIC versions that start with 0xff are
+  IETF drafts. QUIC versions that start with 0x0000 are reserved for IETF
+  consensus documents. For example, QUIC version 1 uses version 0x00000001.
```
mirjak, line 172: Are these convention still true? I guess we don't have that in the transport spec and therefore should probably also remove it here...?

draft-ietf-quic-manageability.md
```diff
@@ -476,26 +483,26 @@ therefore not part of the wire image that is visible to on-path observers.
...
 ## Version Negotiation and Greasing {#version}
 
-Version negotiation is not protected, given the used protection mechanism can
-change with the version.  However, the choices provided in the list of version
-in the Version Negotiation packet will be validated as soon as the cryptographic
-context has been established. Therefore any manipulation of this list will be
-detected and will cause the endpoints to terminate the connection.
+Version Negotiation packets are not intrinsically protected, but QUIC versions
+can use later encrypted messages to verify that they were authentic.
+Therefore any manipulation of this list will be detected and will cause the
```
mirjak, line 488:
```suggestion
Therefore any manipulation of this list will be detected and may cause the
```
as we don't know what future version will do...

draft-ietf-quic-manageability.md
```diff
@@ -526,7 +534,7 @@ provides one bit of information and is quite prone to collide with
 UDP-based protocols other than those that this static bit is meant to allow
 multiplexing with. Second, this feature of the wire image is not invariant
 {{QUIC-INVARIANTS}} and may change in future versions of the protocol, or
-even be negotiated after handshake via future transport parameters.
+even be negotiated during the handshake via proposed transport parameters.
```
mirjak, line 537:
```suggestion
even be negotiated during the handshake via the use of transport parameters.
```

draft-ietf-quic-manageability.md
```diff
@@ -557,19 +565,21 @@ future versions of the protocol.
...
+the network, 0-RTT Protected data packets could be seen before the Initial
+packet.
 
 Note that clients send Initial packets before servers do, servers send Handshake
-packets before  clients do, and only clients send Initial packets with tokens,
-so the sides of a connection can be generally be confirmed by an on-path
-observer. An attempted connection after Retry can be detected by correlating the
-token on the Retry with the token on the subsequent Initial packet.
+packets before clients do, and only clients send Initial packets with tokens.
+Therefore, identity as a client or server can generally be confirmed by an on-
```
mirjak, line 579:
```suggestion
Therefore, the role as a client or server can generally be confirmed by an on-
```

mirjak (changes requested):
Sorry, I have a few more things I just found (not all related to your edits tough). But hopefully easy to fix/decide.

draft-ietf-quic-manageability.md
martinduke: How about "Many QUIC versions that start with 0xff"?  This is a statement of fact, though you're right that in the future other stuff could be in this space.

draft-ietf-quic-manageability.md
martinduke: "Transparent QUIC proxies"? You're right that if the client consents to a proxy (e.g. MASQUE) this isn't strictly true.

draft-ietf-quic-manageability.md
martinduke: This is another warning about the grease, of course. I think it makes sense to include, but we have the other warning about grease in Section 3.1. I'll delete the sentence to move things along, because it's not a big deal.

draft-ietf-quic-manageability.md
mirjak: Yep that's probably good to note!

draft-ietf-quic-manageability.md
mirjak: section 3.1? I thought you actually can grease the fixed bit in QUICv1...?

draft-ietf-quic-manageability.md
mirjak: I think the problem is about the term proxy. For me the term mean only that there is some entity between the client and the server. The whole point of this section was rather to say that a QUIC proxy cannot be transparent in a sense that it can terminate a connection without anybody noticing. And yes, I do have masque in mind.

mirjak: Ah now I see that you added the term "transparent QUIC proxy" in the text. Not sure if I want to coin that term here. Maybe we can just move the part about credentials into a separate sentence, e.g.:

```
A proxy can either explicit identify itself as providing a proxy service or may share the TLS credentials to
authenticate as the server and (in some cases) client acting as a front-facing instance for the offering service itself.
```

martinduke:
No, the greasing is draft-thomson-quic-bit-grease. As you said elsewhere,
the v1 standard itself says the fixed bit MUST be fixed.

draft-ietf-quic-manageability.md
mirjak: Okay, actually I think it would be important here to note that this bit can/should not be used to identify QUIC (rather than talking specifically about greasing). How about saying something like: "Even though this bit is defined as fixed in the QUICv1 base spec, this bit is not invariant between versions or could even still be changed in an extension, therefore it can not be used an identifier for QUIC."

PR #156, opened 2020-12-21
reverse order of "no critical interface", fix #138

### LINKED ISSUE #138: Critical streams aren't worth a whole paragraph
I think that the text on [streams that are critical to an application](https://quicwg.org/ops-drafts/draft-ietf-quic-applicability.html#section-4-4) can be removed.

This was an interesting discussion in HTTP/3, but I'm not sure that the applicability statement needs to concern itself with this detail.

britram:
after discussion with @mirjak, we probably do want to keep the information that "all streams are equal" here. 
Reverse the order of the sentences here to not bury the lede.

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -302,12 +302,10 @@ direction without a peer-generated FIN, much like in TCP. However, an endpoint
 can abruptly close either the ingress or egress direction; these actions are
 fully independent of each other.
 
+QUIC does not provide an interface for exceptional handling of any stream.
 If a stream that is critical for an application is closed, the application can
-generate respective error messages on the application layer to inform the
-other end and/or the higher layer, and eventually indicate QUIC to reset
-the connection. QUIC, however, does not need to know which streams are
-critical, and does not provide an interface for exceptional handling of
-any stream.
+generate error messages on the application layer to inform the other end and/or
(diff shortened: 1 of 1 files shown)


mirjak (approved):

draft-ietf-quic-applicability.md
```diff
@@ -302,12 +302,10 @@ direction without a peer-generated FIN, much like in TCP. However, an endpoint
...
 
+QUIC does not provide an interface for exceptional handling of any stream.
 If a stream that is critical for an application is closed, the application can
-generate respective error messages on the application layer to inform the
-other end and/or the higher layer, and eventually indicate QUIC to reset
-the connection. QUIC, however, does not need to know which streams are
-critical, and does not provide an interface for exceptional handling of
-any stream.
+generate error messages on the application layer to inform the other end and/or
+the higher layer, which can eventually indicate QUIC to reset the connection.
```
martinthomson, line 308: "indicate QUIC to" is awkward phrasing.

martinthomson (approved):
This reads better for sure, aside from an odd bit of phrasing.

draft-ietf-quic-applicability.md
britram: that's because there's a to too few there...

britram: or on review, the application can reset the connection using an interface QUIC provides it. so, reworded.

PR #158, opened 2020-12-21
clarify DDoS, remove LB reference, fix #147

### LINKED ISSUE #147: DDoS guidance is inaccurate about QUIC-LB
"QUIC's linkability resistance ensures that a deliberate connection migration is accompanied by a change in the connection ID and necessitate that connection ID aware DDoS defense system must have the same information about connection IDs as the load balancer {{?I-D.ietf-quic-load-balancers}}. This may be complicated where mitigation and load balancing environments are logically separate."

This is insufficient. Even the load balancer is only able to identify the end server, not the connection.

britram:


mirjak (approved):

martinduke (approved):
This PR removes the problems I cited in the issue. Ship it.

I wonder if we should hit the point in the last sentence a little harder. If you want to support connection migration, there's nothing you can do in this space except the Retry Service stuff in QUIC-LB, is there?

martinthomson (approved):

britram:
Merged...

> If you want to support connection migration, there's nothing you can do in this space except the Retry Service stuff in QUIC-LB, is there?

with the neurons I have left on 22 December, this seems broadly true. I'm not sure we should be too prescriptive about load balancing in this doc, though -- in part because I suspect the challenge of handling DDoS attacks using QUIC traffic will lead to divergent innovation in the LB space in the near term...

mirjak:
So if the 5-tuple changes because of NAT, the connection ID might actually not change. So in case of DDoS you might want to look at the connection ID to at cover those cases. For DDoS you usually don't need a hard guarantee to track all flows (if you block a flow incorrectly, it has to reconnect), but where you can track the flow when the 5-tuple changes, it might actually help. That is what this paragraph is/was about.

PR #161, opened 2020-12-21
Please don't drop new versions, fix #145.

### LINKED ISSUE #145: Ask people not to drop new versions?
The VN section suggests that observers not use the version number to detect QUIC.

Would it be futile to ask them to admit QUIC versions they don't recognize rather than drop them?

britram:


martinduke (approved):

martinthomson (approved):
Wishful thinking, but as you say, can't hurt.

PR #164, opened 2020-12-22
Proposal for rewrite of QoS section

### LINKED ISSUE #140: Rewrite QoS and ECMP section to actually talk about DSCP
Discussion in closed PR #134

mirjak:
fixes #140 (hopefully)

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -881,24 +881,21 @@ and ACK manipulation is impossible. Specifically, heuristics attempting to
 distinguish ACK-only packets from payload-carrying packets based on packet size
 are likely to fail, and are emphatically NOT RECOMMENDED.
 
-## DSCP and ECMP
+## Quality of Service handling and ECMP
 
-QUIC assumes that all packets of a QUIC connection or at least with the
-same 5-tuple {dest addr, source addr, protocol, dest port, source port} will
-receive similar network treatment as feedback provided about loss or delay
-of each packet is used as input to the congestion controller. Therefore it is
-not recommended to use different DiffServ Code Points (DSCPs) {{?RFC2475}} for
-packets belonging to the same connection. If differential network treatment,
-e.g. by the use of different DSCPs, is desired, multiple QUIC 
-connections to the same server may be used. However, in general it is
-recommended to minimize the number of QUIC connections to the same serve, as
-otherwise congestion controllers will compete against each other.
+It is expected that any QoS handling in the network, e.g. based on use of 
+DiffServ Code Points (DSCPs) {{?RFC2475}}, as well as Equal-Cost
+Multi-Path (ECMP) routing is applied on a per flow-basis (and not per-packet)
+and as such that all packets belonging to the
+same QUIC connection get uniform treatment. This is important as
+feedback about loss or delay of each packet is used as input to
+the congestion controller. 
 
-Depending of the loss recovery mechanism implemented, QUIC may be more tolerant
-of packet re-ordering than traditional TCP traffic (see
-{{packetnumber}}). However, currently it is still expected that Equal-Cost
-Multi-Path (ECMP) routing is flow-based and all packets belonging to the
-same QUIC connection get uniform treatment.
+Depending of the loss recovery mechanism implemented, QUIC may be
+more tolerant of packet re-ordering than traditional TCP traffic (see
+{{packetnumber}}). However, it cannot be known by the network which exact
+recovery mechanism is used and therefore reordering tolerance should be
+considered as unknown.
 
 # IANA Considerations
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -883,13 +883,13 @@ are likely to fail, and are emphatically NOT RECOMMENDED.
 
 ## Quality of Service handling and ECMP
 
-It is expected that any QoS handling in the network, e.g. based on use of 
+It is expected that any QoS handling in the network, e.g. based on use of
 DiffServ Code Points (DSCPs) {{?RFC2475}}, as well as Equal-Cost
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -881,29 +881,24 @@ and ACK manipulation is impossible. Specifically, heuristics attempting to
...
-network characteristics.
-
-Given that QUIC is more tolerant of packet re-ordering than TCP (see
-{{packetnumber}}), Equal-cost multi-path routing (ECMP) does not necessarily
-need to be flow based.  However, 5-tuple (plus eventually connection ID if
-present) matching is still beneficial for QoS given all packets are handled by
-the same congestion controller.
+receive similar network treatment as feedback provided about loss or delay
+of each packet is used as input to the congestion controller. Therefore it is
+not recommend to use different DiffServ Code Points (DSCPs) {{?RFC2475}} for
```
martinthomson, line 890:
```suggestion
not recommended to use different DiffServ Code Points (DSCPs) {{?RFC2475}} for
```

draft-ietf-quic-manageability.md
```diff
@@ -881,29 +881,24 @@ and ACK manipulation is impossible. Specifically, heuristics attempting to
...
 are likely to fail, and are emphatically NOT RECOMMENDED.
 
-## QoS support and ECMP
+## DSCP and ECMP
 
-\[EDITOR'S NOTE: this is a bit speculative; keep?]
-
-QUIC does not provide any additional information on requirements on Quality of
-Service (QoS) provided from the network. QUIC assumes that all packets with the
+QUIC assumes that all packets of a QUIC connection or at least with the
```
martinthomson, line 886: Does it?

What you might say instead is:
The congestion control and loss recovery mechanisms defined in {{QUIC-RECOVERY}}, which are optional by the way, function best if the network provides similar treatment to every packet.  Marking packets with different DSCP is likely to result in variations in order, delivery rate, and drop rate.  Using ECMP to distribute packets from a single flow across multiple network paths could produce similar effects.   These variations could adversely affect performance.

The improved feedback in QUIC relative to TCP might mean that QUIC is more able to tolerate variations in packet delivery.  There are also opportunities to deploy loss detection and congestion control algorithms that are more aware of these, but these algorithms do not currently exist.

Getting differential treatment might be better served by using independent connections, but this has other effects, such as increased overheads and competing congestion control.

britram: +1 to Martin's suggestion here.

draft-ietf-quic-manageability.md
```diff
@@ -881,29 +881,24 @@ and ACK manipulation is impossible. Specifically, heuristics attempting to
...
+not recommend to use different DiffServ Code Points (DSCPs) {{?RFC2475}} for
+packets belonging to the same connection. If differential network treatment,
+e.g. by the use of different DSCPs, is desired, multiple QUIC 
+connections to the same server may be used. However, in general it is
+recommended to minimize the number of QUIC connections to the same serve, as
+otherwise congestion controllers will compete against each other.
+
+Depending of the loss recovery mechanism implemented, QUIC may be more tolerant
+of packet re-ordering than traditional TCP traffic (see
+{{packetnumber}}). However, currently it is still expected that Equal-Cost
```
britram, line 899: ...expected by implementations of the protocol?

I'd reword slightly:

"However, the common assumption remains that ECMP is 5-tuple-based, and all packets belonging to the same QUIC connection will get the same treatment."

britram (changes requested):
Better than what's there; suggest one edit (and taking Martin's rephrase suggestion)

draft-ietf-quic-manageability.md
mirjak: Happy to adapt the text but the text as proposed by @martinthomson doesn't seem fully correct to me. Two points: first using _some kind of_ congestion control and recovery is not optional and all of these algorithms have to assume that all packets take the same path/get the same treat because otherwise you will need multiple congestion controller instances and that's not some we have in QUICv1. And second, algorithms do exist. TCP RACK is more re-ordering resilience and the time-based threshold in QUIC is as well. My assumption was/is also that they will more likely be implemented/used/deployed in QUIC than in TCP. 

However, we should probably see if we agree what we want to recommend. I think the right thing to recommend is that you should only use one DCSP with one QUIC connection and if you think you need more than one DCSP for your traffic, you should use multiple QUIC. Not sure if @martinthomson agrees to the second part?

I was also trying to specifically focus this section on DCSP and EMCP (and not QoS in general to much - see new section title) because that the kind of questions I heart.

draft-ietf-quic-manageability.md
mirjak: I meant to say:  at time of writing of this document (and the quic spec) that was/is the community assumption. 

Is there a reason why you changed flow-based to 5-tuple-based? I believe flow is more correct as there might be ECMP scheme that take the IP flow label into account.

britram: was trying to be more precise (with 5-tuple), but as it turns out, that's wrong (DSCP, Flow Label in v6), so... no. leave it "flow-based"

draft-ietf-quic-manageability.md
britram: after discussion with Mirja... the key point here is that all packets handled by the same congestion controller state should use the same DSCP (because they assume the same path treatment), and this is true of all congestion controllers. This is a useful insight, but it belongs in applicability.

A corollary of that belongs here: boxes that are Trying To Help by splitting DSCP on flows *really shouldn't do that* on QUIC, not least because they can't get any information from DPI on which packets need which labels.

britram (changes requested):

mirjak:
okay, next try. Moved some text to applicability and rewrote stuff in manageability... please review again...

britram (approved):

PR #166, opened 2020-12-22
Pointers to sec considerations of quic docs and some more text

### LINKED ISSUE #141: Double check security considerations sections
After removing normative language, it occurs to us that there might be other security considerations that we want to highlight from Transport and TLS in Manageability or (especially) Applicability.

mirjak:
fixes #141

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -921,15 +921,14 @@ QUIC traffic inherently involves tradeoffs with the confidentiality of QUIC's
 control information; this entire document is therefore security-relevant.
 
 More security considerations for QUIC are discussed in {{!QUIC-TRANSPORT}}
-and {{!QUIC-TLS}},
-generally considering active or passive attackers in the network as well as attacks
-on specific QUIC mechanism.
+and {{!QUIC-TLS}}, generally considering active or passive attackers in the
+network as well as attacks on specific QUIC mechanism.
 
-Version Negotiation packets do not contain any mechanism to prevent version downgrade
-attacks. However, future versions of QUIC that use Version Negotiation packets are
-require to define a mechanism that is robust against version downgrade attacks.
-Therefore a network node should not attempt to impact version selection, as version
-downgrade may result in connection failure.
+Version Negotiation packets do not contain any mechanism to prevent version
+downgrade attacks. However, future versions of QUIC that use Version Negotiation
+packets are require to define a mechanism that is robust against version
+downgrade attacks. Therefore a network node should not attempt to impact version
+selection, as version downgrade may result in connection failure.
 
 # Contributors
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -911,9 +911,24 @@ This document has no actions for IANA.
 
 # Security Considerations
 
-Supporting manageability of QUIC traffic inherently involves tradeoffs with the
-confidentiality of QUIC's control information; this entire document is therefore
-security-relevant.
+QUIC is an encrypted and authenticated transport. That means, once the
+cryptographic handshake is complete, QUIC endpoints discard most packets that
+are not authenticated, greatly limiting the ability of an attacker to interfere
+with existing connections.
+
+However, some information are still observerable, as supporting manageability of
+QUIC traffic inherently involves tradeoffs with the confidentiality of QUIC's
+control information; this entire document is therefore security-relevant.
+
+More security considerations for QUIC are discussed in {{!QUIC}} and {{!QUIC-TLS}},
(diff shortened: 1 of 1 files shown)


mirjak:
This is a proposal to mention some more things specifically in the security considerations section. At minimum we should add a sentence pointing the reader at the security consideration of the transport and tis drafts. I additionally, more or less, copied some notes from the transport draft that I found particular relevant for reader of the manageability draft. Comments welcome!

draft-ietf-quic-manageability.md
```diff
@@ -911,9 +911,24 @@ This document has no actions for IANA.
...
+
+More security considerations for QUIC are discussed in {{!QUIC}} and {{!QUIC-TLS}},
+generally considering active or passive attackers in the network as well as attacks
+on specific QUIC mechanism.
+
+Version Negotiation packets do not contain any mechanism to prevent version downgrade
+attacks. However, future versions of QUIC that use Version Negotiation packets are
+require to define a mechanism that is robust against version downgrade attacks.
+Therefore a network node can and should not attempt to impact version selection and
+as such version downgrade is considered an attack.
```
martinthomson, line 931: There are a number of active attacks against a QUIC handshake.  All of them result in connection failures.  This will be no different.  I don't really see the point of isolating this, except to note that the specific methods involved are not yet defined.  Dropping this last sentence would be a start, but I would reconsider the entire paragraph.

britram:
```suggestion
Therefore a network node should not attempt to impact version selection, as version
downgrade may result in connection failure.
```

draft-ietf-quic-manageability.md
```diff
@@ -911,9 +911,24 @@ This document has no actions for IANA.
 
 # Security Considerations
 
-Supporting manageability of QUIC traffic inherently involves tradeoffs with the
-confidentiality of QUIC's control information; this entire document is therefore
-security-relevant.
+QUIC is an encrypted and authenticated transport. That means, once the
+cryptographic handshake is complete, QUIC endpoints discard most packets that
+are not authenticated, greatly limiting the ability of an attacker to interfere
+with existing connections.
```
britram, line 917: Perhaps note that on-path modifications of packets will result in one of two outcomes:

- packet ignored (and frames eventually retransmitted, if necessary)
- connection dropped

draft-ietf-quic-manageability.md
mirjak: I wanted to explicitly about version downgrades because I was thinking that the people who do version filtering might easily get the idea that version downgrade could be a solution to some of their problems and would not think of it as an attack. I agree that this is true for any handshake "attack" but I would like to be as explicit as possible about what we recommend and not recommend in this doc for cases where we can already foresee problems.

mirjak:
@martinthomson any further thought here?

draft-ietf-quic-manageability.md
```diff
@@ -911,9 +911,24 @@ This document has no actions for IANA.
...
+
+However, some information are still observerable, as supporting manageability of
```
britram, line 919:
```suggestion
However, some information is still observerable, as supporting manageability of
```

draft-ietf-quic-manageability.md
```diff
@@ -911,9 +911,24 @@ This document has no actions for IANA.
...
+QUIC is an encrypted and authenticated transport. That means, once the
+cryptographic handshake is complete, QUIC endpoints discard most packets that
+are not authenticated, greatly limiting the ability of an attacker to interfere
+with existing connections.
+
+However, some information is still observerable, as supporting manageability of
+QUIC traffic inherently involves tradeoffs with the confidentiality of QUIC's
+control information; this entire document is therefore security-relevant.
+
+More security considerations for QUIC are discussed in {{!QUIC}} and {{!QUIC-TLS}},
```
britram, line 923:
```suggestion
More security considerations for QUIC are discussed in {{!QUIC-TRANSPORT}}
and {{!QUIC-TLS}},
```

draft-ietf-quic-manageability.md
```diff
@@ -911,9 +911,25 @@ This document has no actions for IANA.
...
+are not authenticated, greatly limiting the ability of an attacker to interfere
+with existing connections.
+
+However, some information is still observerable, as supporting manageability of
+QUIC traffic inherently involves tradeoffs with the confidentiality of QUIC's
+control information; this entire document is therefore security-relevant.
+
+More security considerations for QUIC are discussed in {{!QUIC-TRANSPORT}}
+and {{!QUIC-TLS}},
+generally considering active or passive attackers in the network as well as attacks
```
britram, line 925:
```suggestion
generally considering active or passive attackers in the network as well
as attacks
```

PR #168, opened 2020-12-22
Do we want to add a reference to the HTTP/3 draft as well

mirjak:
to the security considerations section?

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -713,9 +713,8 @@ properties as QUIC; if this is not possible, the connection should fail to
 allow the application to explicitly handle fallback to a less-secure
 alternative. See {{fallback}}.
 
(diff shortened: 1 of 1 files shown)


mirjak:
Cross protocol attacks, traffic analysis, and migration are discussed as well in the transport draft, so this pointer might not be absolutely necessary. However, we might still want to have it and then I think these are the relevant sections to refer to.

britram:
Applicability to me seems to be mostly targeted to non-H3 users of QUIC, so I'm not sure the H3-specific considerations need to be mentioned here.

mirjak:
My thinking was to add that as reference for other applications that might need to consider similar attacks/security issues. At least half of the points in the hhtp/3 sec considerations are really http specific and probably don't have value for other applications but there are a number of points which are more general. Those points are to some extend probably also covered by the transport draft but may be explained in more detail in the h3 draft.

britram (approved):

PR #170, opened 2021-01-12
Add Retry discussion to DDoS section

### LINKED ISSUE #148: Should DDoS section talk about Retry Services?
I am not really sure at all that the existing DDoS guidance is good advice; we have a somewhat more robust plan in quic-lb, which perhaps we should reference?

martinduke:
Fix #148


britram (approved):

mirjak:
The text is good however I don't understand its placement in the draft fully. Shouldn't this paragraph go into the previous section or even in an own section?

I know the issue, that this PR is addressing, was about DDoS but if we want to have it in the DDoS section we need probably some more words explaining how Retry should be used in this case. Currently it comes a bit out of nothing and interrupts the reading flow of the rest of the text...

PR #207, opened 2021-02-08
remove NOT RECOMMENDED in ACK section

### LINKED ISSUE #182: Sect 4.6: emphatically NOT RECOMMENDED use of ACK mitigation?
The text says: “and are emphatically NOT RECOMMENDED.” I think I would agree with a strong encouragement to not do this, but I would suggest to not use upper case letters at this point, and I’m not sure “emphatically” is a great word in a RFC :-).

I’d also note that the volume of ACK traffic really can be a problem that QUIC needs to address, because QUIC over radio (of various types) can be significantly more expensive or lower performant than TCP, precisely because the control/ACK traffic is much more and there is no opportunity for network mitigation - hence it behoves QUIC implementors to use appropriate methods.

mirjak:
fixes #182

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -894,7 +894,7 @@ QUIC, since acknowledgment signaling is carried inside QUIC's encrypted payload,
 and ACK manipulation is impossible. Specifically, heuristics attempting to
 distinguish ACK-only packets from payload-carrying packets based on packet size
 are likely to fail, and are not recommended to use as a way to construe
-internals of QUIC's operator as those mechanisms can change, e.g., due to the
+internals of QUIC's operation as those mechanisms can change, e.g., due to the
 use of extensions.
 
 ## Quality of Service handling and ECMP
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -893,7 +893,9 @@ signaling. Distinguishing ACK packets is trivial in TCP, but not supported by
 QUIC, since acknowledgment signaling is carried inside QUIC's encrypted payload,
 and ACK manipulation is impossible. Specifically, heuristics attempting to
 distinguish ACK-only packets from payload-carrying packets based on packet size
-are likely to fail, and are emphatically NOT RECOMMENDED.
+are likely to fail, and are not recommended to use as a way to construe
+internals of QUIC's operator as those mechanisms can change, e.g., due to the
```
gorryfair, line 897: /operator/operation/

seanturner (approved):

draft-ietf-quic-manageability.md
mirjak, line 897:
```suggestion
internals of QUIC's operation as those mechanisms can change, e.g., due to the
```

PR #208, opened 2021-02-08
Cite RFC4748 correctly

mirjak:


draft-ietf-quic-manageability.md
```diff
@@ -790,12 +790,13 @@ signal ({{sec-teardown}}) means that this state must be purged either through
 timers or through least-recently-used eviction, depending on application
 requirements.
 
-{{?RFC4787}} recommends a 2 minute timeout interval for UDP. However,
-timers can be lower, in the range of 15 to 30 seconds. In contrast,
-{{?RFC5382}} recommends a timeout of more than 2 hours for TCP, given that TCP
-is a connection-oriented protocol with well-defined closure semantics. For
-network devices that are QUIC-aware, it is recommended to also use
-longer timeouts for QUIC traffic, as QUIC is connection-oriented. As such,
+{{?RFC4787}} requires a timout that is not less than 2 minutes for most UDP
```
martinthomson, line 793:
```suggestion
{{?RFC4787}} requires a timout that is at least than 2 minutes for most UDP
```

mirjak: Happy to use your wording but I actually used the wording form RFC4787. Actually the wording is

" A NAT UDP mapping timer MUST NOT expire in less than two minutes, unless REQ-5a applies."

gorryfair: And I WOUD think it helpful to state here why 2 minutes is better than shorter values, so that the read knows we are serious.

mirjak: Yep, @gorryfair  that will go into a separate PR! Issues are still open to address this!

PR #212, opened 2021-02-08
Add proposed section on guiding Path MTU

### LINKED ISSUE #210: Network management of Path MTU
The manageability draft aims to provide guidance to network operators in managing QUIC traffic, yet currently says nothing about what could be a major impact for some networks.

Consider a cellular/mobile network that is providing internet access to end users. Because it is an internet access service, the users have an implicit expectation that the complete range of internet packet sizes (i.e. up to 1500 bytes) can be exchanged. And as it is a cellular network packets are routinely encapsulated in GTP-U. The network operator also chooses to add IPSEC encapsulation in order to secure communications over a diverse range of connectivity providers. Since at least some of that connectivity is over paths with 1500 MTU, it implies that end-user datagrams over a certain size (e.g. 1400) would have to be fragmented at the point of encapsulation, and reassembled before delivery to the user.

Such fragmentation and reassembly comes at a cost. Importantly it implies a limited capacity for packets with sizes in the relevant range (e.g. 1400-1500). In a world where TCP dominates this is easily managed with MSS Clamping, so that both TCP endpoints learn the optimal size to use. In QUIC a different mechanism is required. QUIC-TRANSPORT describes how the endpoints can discover a hard path MTU limit, i.e. one where there is a clear cutoff above a certain size. I think the manageability draft needs a section that advises operators of networks where there is a soft path MTU limit. I'll create a PR with proposed wording for this section.

If anyone is unconvinced about the need for such packet size limitation, happy to discuss in more depth.

chris-box:
This is my suggestion to resolve #210.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -1051,13 +1051,15 @@ PMTU information between IP flows, in the IP-layer cache, so short-term
 consistency between the PMTU for flows can help avoid an endpoint using a
 PMTU that is inefficient.
 
-If network operators wish to start to drop large packets to avoid
-fragmentation, they may wish to focus on QUIC. QUIC packets cannot always
-be identified ({{sec-identifying}}), but we assume that a set of candidate
-QUIC flows have been chosen. For QUIC endpoints using DPLPMTUD it is
-minimally sufficient for the path to silently drop a packet larger than the
-supported size. (A QUIC probe packet is used to discover the PMTU. If lost,
-this does not impact the flow of QUIC data.)
+Networks with configurations that would lead to fragmentation of large packets
+should drop such packets rather than fragmenting them. Network operators who
+plan to implement a more selective policy may start by focussing on QUIC.
+QUIC flows cannot always be easily distinguished from other UDP traffic, but
+we assume at least some portion of QUIC traffic can be identified
+(see {{sec-identifying}}). For QUIC endpoints using DPLPMTUD it is recommended
+for the path to drop a packet larger than the supported size. A QUIC probe
+packet is used to discover the PMTU. If lost, this does not impact the flow of
+QUIC data.
 
 IPv4 routers generate an ICMP message when a packet is dropped because the
 link MTU was exceeded. {{RFC8504}} specifies how an IPv6 node generates an
@@ -1070,7 +1072,9 @@ Since a network cannot know in advance which discovery method a QUIC endpoint
 is using, it should always send a PTB message in addition to dropping the
 oversized packet. A generated PTB message should be compliant with the
 validation requirements of {{Section 14.2.1 of QUIC-TRANSPORT}}, otherwise it
-will be ignored by DPLPMTUD.
+will be ignored by DPLPMTUD. This will likely provide the right signal for the
+endpoint to keep the packet size small and thereby avoid network fragmentation
+for that flow entirely.
 
 # IANA Considerations
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -1025,13 +1025,13 @@ result in slow and costly timeouts before abandoning the connection.
 
 ## Guiding Path MTU
 
-Some networks are required to be able to carry 1500-byte packets but become
-inefficient at this size, because such packets are fragmented and
-reassembled across a segment where the effective MTU is smaller (e.g. due to
-IPSEC overhead). They do this even when the encapsulated packet is not a
-candidate for fragmentation by routers (it's IPv6, or has DF bit set). In such
-networks this leads to a need to influence a majority of senders to use smaller
-packets, so that the limited reassembly capacity is not exceeded.
+Some networks support 1500-byte packets, but can only do so by fragmenting at a
+lower layer before traversing a smaller MTU segment, and then reassembling.
+This is permissible even when the IP layer is IPv6 or IPv4 with the DF bit set,
+because it occurs below the IP layer. However, this process can add to compute
+and memory costs, leading to a bottleneck that limits network capacity. In such
+networks this generates a desire to influence a majority of senders to use
+smaller packets, so that the limited reassembly capacity is not exceeded.
 
 For TCP, MSS clamping (Section 3.2 of {{?RFC4459}}) is often used to change
 the sender's maximum TCP segment size, but QUIC requires a different approach.
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -1046,7 +1046,7 @@ cause impact to the flow of data. It serves only to guide the discovery of an
 optimal Path MTU.
 
 DPLPMTUD can optionally make use of ICMP Packet Too Big messages (PTB, defined
-in section 1.1 of {{DPLPMTUD}}). PMTUD requires such PTB messages in order to
+in {{Section 1.1 of DPLPMTUD}}). PMTUD requires such PTB messages in order to
 work. Since the network cannot know in advance which discovery method the QUIC
 endpoints are using, it should send a PTB packet in addition to dropping the
 oversized probe packet. The generated PTB packet should be compliant with the
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -1023,6 +1023,36 @@ not arrive at their destination. Admitting a few packets allows the QUIC
 endpoint to determine that the path accepts QUIC. Sudden drops afterwards will
 result in slow and costly timeouts before abandoning the connection.
 
+## Guiding Path MTU
+
+Some networks are required to be able to carry 1500-byte packets but become
+inefficient at this size, because such packets require fragmentation and
+reassembly across a segment where the effective MTU is smaller (e.g. due to
+IPSEC overhead). This leads to a need to influence a majority of senders to
+use smaller packets, so that the limited reassembly capacity is not exceeded.
+
+If the majority use TCP this can be managed with MSS Clamping, but QUIC
+requires a different approach. Section 14 of {{QUIC-TRANSPORT}} advises senders
+to probe larger sizes using Datagram Packetization Layer PMTU Discovery
+({{?DPLPMTUD=RFC8899}}) or Path Maximum Transmission Unit Discovery
+(PMTUD: {{?RFC1191}} and {{?RFC8201}}). This mechanism will encourage senders to
+approach the maximum size, driving fragmentation that they may not be aware of.
+
+Networks can seek to avoid this by identifying probable QUIC packets
+({{sec-identifying}}) and taking action whenever packets exceed the desired
+size. For DPLPMTUD a sufficient action is to drop the packet. If it is
+genuinely QUIC, the packet should have been a probe and so its loss should not
+cause impact to the flow of data. It serves only to guide the discovery of an
+optimal Path MTU.
+
+DPLPMTUD can optionally make use of ICMP Packet Too Big messages (PTB, defined
+in section 1.1 of {{DPLPMTUD}}). PMTUD requires such PTB messages in order to
+work. Since the network cannot know in advance which discovery method the QUIC
+endpoints are using, it should send a PTB packet in addition to dropping the
+oversized probe packet. The generated PTB packet should be compliant with the
+validation requirements of section 14.2.1 of {{QUIC-TRANSPORT}} otherwise it
+is likely to be ignored.
+
 # IANA Considerations
 
 This document has no actions for IANA.
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -1041,7 +1041,7 @@ the sender's maximum TCP segment size, but QUIC requires a different approach.
 Datagram Packetization Layer PMTU Discovery ({{?DPLPMTUD=RFC8899}}) or Path
 Maximum Transmission Unit Discovery (PMTUD: {{?RFC1191}} and {{?RFC8201}}).
 This mechanism will encourage senders to approach the maximum size, which
-could cause fragmentation that they may not be aware of.
+could cause fragmentation with a network segment that they may not be aware of.
 
 If path performance is limited when sending larger packets, an on-path
 device should support a maximum packet size for a specific transport flow
@@ -1051,8 +1051,8 @@ PMTU information between IP flows, in the IP-layer cache, so short-term
 consistency between the PMTU for flows can help avoid an endpoint using a
 PMTU that is inefficient.
 
-If network operators wish to start with a more limited implementation of
-such packet drops, they may wish to focus on QUIC. QUIC packets cannot always
+If network operators wish to start to drop large packets to avoid
+fragmentation, they may wish to focus on QUIC. QUIC packets cannot always
 be identified ({{sec-identifying}}), but we assume that a set of candidate
 QUIC flows have been chosen. For QUIC endpoints using DPLPMTUD it is
 minimally sufficient for the path to silently drop a packet larger than the
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -1038,8 +1038,8 @@ the sender's maximum TCP segment size, but QUIC requires a different approach.
 {{Section 14 of QUIC-TRANSPORT}} advises senders to probe larger sizes using
 Datagram Packetization Layer PMTU Discovery ({{?DPLPMTUD=RFC8899}}) or Path
 Maximum Transmission Unit Discovery (PMTUD: {{?RFC1191}} and {{?RFC8201}}).
-This mechanism will encourage senders to approach the maximum size, driving
-fragmentation that they may not be aware of.
+This mechanism will encourage senders to approach the maximum size, which
+could cause fragmentation that they may not be aware of.
 
 If path performance is limited when sending larger packets, an on-path
 device should support a maximum packet size for a specific transport flow
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -1033,7 +1033,7 @@ candidate for fragmentation by routers (it's IPv6, or has DF bit set). In such
 networks this leads to a need to influence a majority of senders to use smaller
 packets, so that the limited reassembly capacity is not exceeded.
 
-For TCP, MSS clamping ({{?Section 3.2 of RFC4459}}) is often used to change
+For TCP, MSS clamping (Section 3.2 of {{?RFC4459}}) is often used to change
 the sender's maximum TCP segment size, but QUIC requires a different approach.
 Section 14 of {{QUIC-TRANSPORT}} advises senders to probe larger sizes using
 Datagram Packetization Layer PMTU Discovery ({{?DPLPMTUD=RFC8899}}) or Path
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -60,6 +60,8 @@ informative:
       -
         ins: N. Brownlee
     date: 2014-04
+  RFC4459:
+  RFC8504:
 
 --- abstract
 
@@ -1033,7 +1035,7 @@ and memory costs, leading to a bottleneck that limits network capacity. In such
 networks this generates a desire to influence a majority of senders to use
 smaller packets, so that the limited reassembly capacity is not exceeded.
 
-For TCP, MSS clamping (Section 3.2 of {{?RFC4459}}) is often used to change
+For TCP, MSS clamping ({{Section 3.2 of RFC4459}}) is often used to change
 the sender's maximum TCP segment size, but QUIC requires a different approach.
 {{Section 14 of QUIC-TRANSPORT}} advises senders to probe larger sizes using
 Datagram Packetization Layer PMTU Discovery ({{?DPLPMTUD=RFC8899}}) or Path
@@ -1067,7 +1069,7 @@ performance {{Section 4.6 of DPLPMTUD}}.
 Since a network cannot know in advance which discovery method a QUIC endpoint
 is using, it should always send a PTB message in addition to dropping the
 oversized packet. A generated PTB message should be compliant with the
-validation requirements of section 14.2.1 of {{QUIC-TRANSPORT}}, otherwise it
+validation requirements of {{Section 14.2.1 of QUIC-TRANSPORT}}, otherwise it
 will be ignored by DPLPMTUD.
 
 # IANA Considerations
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -1035,7 +1035,7 @@ smaller packets, so that the limited reassembly capacity is not exceeded.
 
 For TCP, MSS clamping (Section 3.2 of {{?RFC4459}}) is often used to change
 the sender's maximum TCP segment size, but QUIC requires a different approach.
-Section 14 of {{QUIC-TRANSPORT}} advises senders to probe larger sizes using
+{{Section 14 of QUIC-TRANSPORT}} advises senders to probe larger sizes using
 Datagram Packetization Layer PMTU Discovery ({{?DPLPMTUD=RFC8899}}) or Path
 Maximum Transmission Unit Discovery (PMTUD: {{?RFC1191}} and {{?RFC8201}}).
 This mechanism will encourage senders to approach the maximum size, driving
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -1041,20 +1041,34 @@ Maximum Transmission Unit Discovery (PMTUD: {{?RFC1191}} and {{?RFC8201}}).
 This mechanism will encourage senders to approach the maximum size, driving
 fragmentation that they may not be aware of.
 
-Networks can seek to avoid this by identifying probable QUIC packets
-({{sec-identifying}}) and taking action whenever packets exceed the desired
-size. For DPLPMTUD a sufficient action is to drop the packet. If it is
-genuinely QUIC, the packet should have been a probe and so its loss should not
-cause impact to the flow of data. It serves only to guide the discovery of an
-optimal Path MTU.
+If path performance is limited when sending larger packets, an on-path
+device should support a maximum packet size for a specific transport flow
+and then consistently drop all packets that exceed the configured size
+when the inner IPv4 packet has DF set, or IPv6 is used. Endpoints can cache
+PMTU information between IP flows, in the IP-layer cache, so short-term
+consistency between the PMTU for flows can help avoid an endpoint using a
+PMTU that is inefficient.
 
-DPLPMTUD can optionally make use of ICMP Packet Too Big messages (PTB, defined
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -1023,6 +1023,36 @@ not arrive at their destination. Admitting a few packets allows the QUIC
...
+use smaller packets, so that the limited reassembly capacity is not exceeded.
+
+If the majority use TCP this can be managed with MSS Clamping, but QUIC
+requires a different approach. Section 14 of {{QUIC-TRANSPORT}} advises senders
+to probe larger sizes using Datagram Packetization Layer PMTU Discovery
+({{?DPLPMTUD=RFC8899}}) or Path Maximum Transmission Unit Discovery
+(PMTUD: {{?RFC1191}} and {{?RFC8201}}). This mechanism will encourage senders to
+approach the maximum size, driving fragmentation that they may not be aware of.
+
+Networks can seek to avoid this by identifying probable QUIC packets
```
mirjak, line 1041: The problem is that probe packets cannot be identified/distinguished from other QUIc traffic. I believe that's also what section 3.1 (Identifying QUIC traffic) says. If that's no clear we can say more there.

chris-box: My interpretation of quic-transport is that it says:
- Initial packets would not normally exceed 1200 bytes unless the client is aware (from history) that a larger size is supported.
- Endpoints should only send packets larger than this if they are employing either PMTUD or DPLPMTUD.
- All QUIC packets that are not sent in a PMTU probe should be sized to fit within the (currently known) maximum datagram size.

Hence probes should be identifiable as the only packets that are larger than this. Is this wrong?

draft-ietf-quic-manageability.md
```diff
@@ -1023,6 +1023,36 @@ not arrive at their destination. Admitting a few packets allows the QUIC
 endpoint to determine that the path accepts QUIC. Sudden drops afterwards will
 result in slow and costly timeouts before abandoning the connection.
 
+## Guiding Path MTU
+
+Some networks are required to be able to carry 1500-byte packets but become
+inefficient at this size, because such packets require fragmentation and
+reassembly across a segment where the effective MTU is smaller (e.g. due to
+IPSEC overhead). This leads to a need to influence a majority of senders to
+use smaller packets, so that the limited reassembly capacity is not exceeded.
```
ianswett, line 1032: It's worth calling out that these networks fragment instead of dropping v4 packets with DF set and IPv6 packets(assuming that's what they do)

ianswett (commented):
Thanks for bringing this issue up, as we hit this in production and were quite surprised.

draft-ietf-quic-manageability.md
```diff
@@ -1023,6 +1023,36 @@ not arrive at their destination. Admitting a few packets allows the QUIC
...
+
+Networks can seek to avoid this by identifying probable QUIC packets
+({{sec-identifying}}) and taking action whenever packets exceed the desired
+size. For DPLPMTUD a sufficient action is to drop the packet. If it is
+genuinely QUIC, the packet should have been a probe and so its loss should not
+cause impact to the flow of data. It serves only to guide the discovery of an
+optimal Path MTU.
+
+DPLPMTUD can optionally make use of ICMP Packet Too Big messages (PTB, defined
+in section 1.1 of {{DPLPMTUD}}). PMTUD requires such PTB messages in order to
```
martinthomson, line 1049:
```suggestion
in {{Section 1.1 of DPLPMTUD}}). PMTUD requires such PTB messages in order to
```

This reference format is relatively new, but also really nice.

martinthomson (commented):
I think that this needs to be clearer: you are proposing that networks a) identify QUIC, and b) identify large QUIC packets.  Based on this, you are recommending that all of these large packets be dropped rather than fragmented.

As this is something that networks are always allowed to do - networks are entitled to decide their own MTU - this is OK.  However, I'm wondering why you need to condition this on the protocol being QUIC.

draft-ietf-quic-manageability.md
chris-box: Ian that's interesting. The relevant sentence for me from quic-transport is "Datagrams containing Initial packets MAY exceed 1200 bytes if the sender believes that the network path and peer both support the size that it chooses." In the general case, from where does that belief arise? In starting at 1330 (which I assume is the v6 size) it would work fine on the EE network but it would fail in traversing a network with the IPv6 minimum MTU. Are you relying on successful PMTU discovery from the INITIAL packet to catch that situation?

draft-ietf-quic-manageability.md
chris-box: Yes happy to make this clearer in the text.

chris-box:
> 
> 
> I think that this needs to be clearer: you are proposing that networks a) identify QUIC, and b) identify large QUIC packets. Based on this, you are recommending that all of these large packets be dropped rather than fragmented.
> 
> As this is something that networks are always allowed to do - networks are entitled to decide their own MTU - this is OK. However, I'm wondering why you need to condition this on the protocol being QUIC.

Good question. I think it'd be unwise to take this action (drop & send ICMP) blindly without regard to how the transport protocol will react. UDP of course carries a diverse set of services, and not all of them will react well to this. If all implementations complied with RFCs there should be no issue; a DF bit set means the sender is expecting it might have to deal with drops. Hopefully they have tested that it works gracefully.

But from a sense of caution I prefer to take the approach that such actions should be targetted on services where we have more confidence that it will react well. The finalised QUIC is one such transport, and significantly it's one that will carry the majority of internet traffic before long.

draft-ietf-quic-manageability.md
ianswett: HTTP/3 has to handle fallback to TCP no matter what, and too large a packet and UDP being dropped look identical to a client attempting a handshake.

In this case, the belief that networks support 1330/1350 byte size is based on historical data we gathered showing what MTUs were supported by most networks, so it's not path specific.

mjoras:
> As this is something that networks are always allowed to do - networks are entitled to decide their own MTU - this is OK. However, I'm wondering why you need to condition this on the protocol being QUIC.

I find @chris-box's reasoning pretty convincing.

The on path equipment is already taking different approaches based on the transport protocol (i.e. MSS clamping with TCP) and allowing some existing protocols to be fragmented. QUIC being treated as the latter category is obviously problematic, so the operator has two choices that would allow DPLPMTUD to function for QUIC:

1. Stop doing any sort of fragmentation.
2. Make a best effort attempt to not fragment QUIC packets and rather drop them on the floor. 

The former is likely to disrupt some protocols which are functioning "fine" today with fragmentation and reassembly, which may make it untenable for an operator. The latter, while not totally straightforward and smells of ossification, seems like it would be more tenable in practice while addressing the actual issue.

ianswett (approved):
I think this is enough to make the problem and proposed solution clear.  Thanks for writing it.

mirjak:
What is still not clear to me is why is the advise here  "please drop big QUIC packet (rather than fragment)" instead of directly saying "please drop if DF is set" (which I think is the advise we already have in various RFCs). If operators don't follow the later advise now, why should they follow the former?

chris-box:
Mirja, it is true that "please drop if DF is set, and send PTB" is existing advice that is not universally obeyed. And the advice in this PR will not be universally obeyed either, but some operators will take note, and some developers of PGW software will write code to implement it. (This is not a natural function for a PGW, because the MTU on the S1-U side is large. It has to be deliberately configured based on knowledge of the downstream network, just like MSS clamping currently is.)

You're right to ask why should the implementations be focussed on QUIC, rather than what it should be from an architectural purity perspective, which is all IP that has a DF property. Here my answer is that any network operator contemplating deliberately preventing some traffic that they current permit will be naturally cautious. They will not want unintended consequences, and not want to be on the receiving end of complaints about an obscure service that worked yesterday but doesn't today. Hence it is a much safer bet to focus the intervention on the single, well-understood, major traffic type that is driving the change.

draft-ietf-quic-manageability.md
```diff
@@ -1023,6 +1023,38 @@ not arrive at their destination. Admitting a few packets allows the QUIC
...
+## Guiding Path MTU
+
+Some networks are required to be able to carry 1500-byte packets but become
+inefficient at this size, because such packets are fragmented and
+reassembled across a segment where the effective MTU is smaller (e.g. due to
+IPSEC overhead). They do this even when the encapsulated packet is not a
+candidate for fragmentation by routers (it's IPv6, or has DF bit set). In such
+networks this leads to a need to influence a majority of senders to use smaller
+packets, so that the limited reassembly capacity is not exceeded.
+
```
gorryfair, line 1035: This makes it sound like the current method is very wrong - It is **OK* for a lower layer (including a tunnel) to fragment and reassemble to the MTU it supports. However, in some cases it **CAN** fragmentation and reassembly below IP can add to the processing cost and memory requirements - and hence can become a bottleneck. This text isn't clear yet  - but the text might be heading in the correct direction

draft-ietf-quic-manageability.md
```diff
@@ -1023,6 +1023,38 @@ not arrive at their destination. Admitting a few packets allows the QUIC
...
+If the majority use TCP this can be managed with MSS Clamping, but QUIC
+requires a different approach. Section 14 of {{QUIC-TRANSPORT}} advises senders
```
gorryfair, line 1037: /majority use/majority of traffic use/
... although I really do think it better to avoid /can/ and say /is/... 

I suggest:

/For TCP, MSS-clamping is currently used to change the sender's maximum TCP segment size, but QUIC.../

draft-ietf-quic-manageability.md
gorryfair: I believe the recommendation needs to be that path needs to support the minimum QUIC packet size (aka 1200B).  If the path performance is limited when sending larger packets, the devices on-path need to decide consistently send all packets for a flow, or should drop all packets that exceed a specific size. This avoids creating an anomaly that has unpredictable consequences on QUIC protocol mechanisms that probe a path. ((The same predictability is needing when choosing a forwarding path, etc)).

draft-ietf-quic-manageability.md
```diff
@@ -1023,6 +1023,38 @@ not arrive at their destination. Admitting a few packets allows the QUIC
...
+approach the maximum size, driving fragmentation that they may not be aware of.
+
+Networks can seek to avoid this by identifying probable QUIC packets
+({{sec-identifying}}) and taking action whenever packets exceed the desired
+size. For DPLPMTUD a sufficient action is to drop the packet. If it is
+genuinely QUIC, the packet should have been a probe and so its loss should not
+cause impact to the flow of data. It serves only to guide the discovery of an
+optimal Path MTU.
+
+DPLPMTUD can optionally make use of ICMP Packet Too Big messages (PTB, defined
```
gorryfair, line 1050: When a device drops an IP packet, it ought to send an ICMP PTB message indicating the supported MTU, and quoting the discarded packet. However, endpoints do not necessarily receive these PTB Messages, and QUIC is robust to their loss.

gorryfair (commented):
This is heading in the direction I expected, but would benefit from more polishing.

draft-ietf-quic-manageability.md
chris-box: Would it be better to say this?

"Some networks are required to support 1500-byte packets but can only do so by fragmenting at a lower layer before traversing a smaller MTU segment, and then reassembling. This is permissible even when the IP layer is IPv6 or has DF bit set, because it occurs at that lower layer. However this process can add to compute and memory costs, leading to a bottleneck that limits network capacity. In such networks this generates a desire to influence a majority of senders to use smaller packets, so that the limited reassembly capacity is not exceeded."

draft-ietf-quic-manageability.md
chris-box: Happy with the new suggestion, so I'll merge it. I think I should also cite RFC4459 here.

draft-ietf-quic-manageability.md
chris-box: Agree that "If the path performance is limited when sending larger packets, the devices on-path need to decide consistently send all packets for a flow, or should drop all packets that exceed a specific size. "
This is however a very limited consistency (one flow). I feel we should go further and ask for consistency for all flows of a particular transport protocol (QUIC in this case). This would permit endpoints to make use of historic knowledge such as "on this network, QUIC only works up to size x".

We should also state the desired goal that it should be consistent for all IP packets, regardless of protocol, but recognise that network may be cautious here, especially when there is a large diversity of protocols.

draft-ietf-quic-manageability.md
chris-box: Agree. Are you suggesting the wording should be tightened, so that dropping the packet and not bothering with the PTB is called out as less-desirable?

draft-ietf-quic-manageability.md
chris-box: https://github.com/chris-box/ops-drafts/commit/4dbf7d8137ab17dc7de54644c2df18e6c172d380

draft-ietf-quic-manageability.md
gorryfair: Now I see where this is heading, I think we should encourage PTB ... and perhaps say something like:

"({{sec-identifying}}) and taking action whenever packets exceed the desired size. For DPLPMTUD it is sufficient for the path to silently drop a packet larger than the supported size. (A QUIC probe packet is used to discover the PMTU. If lost, this does not impact the flow of QUIC data.) 

IPv4 routers generate an ICMP message when a packet is dropped because the link MTU was exceeded. {{RFC8504}} specifies how an IPv6 node generates an ICMPv6 Packet Too Big message (PTB) in this case. 

PMTUD relies upon an endpoint receiving such PTB messages {{RFC8201}},  whereas DPLPMTUD does not reply upon these messages, but still can optionally utilise these to improve performance {{Section 4.6 of DPLPMTUD}}. 

Since a network cannot know in advance which discovery method a QUIC endpoint is using, it ought to send a PTB message in addition to dropping any oversized packet.  A generated PTB message should be compliant with the validation requirements of {{section 14.2.1 of QUIC-TRANSPORT}}, otherwise it will be ignored by DPLPMTUD."

- choose: /ought to send a PTB message/needs to send/ - choose appropriate words!

gorryfair: Details:
- Unwelcome detail: In some cases, a PMTUD implementation *might* accept any ICMP  message to update the IP cache, but that isn't a great choice, and text in recent RFCs seeks to discourage this.
- As far as I recall, the RFC series is very murky for IPv4. 
- For IPv6 the advice is clear, despite the current state of affairs reported in RFC8899, and that this alone can't really solve the problem. It says: "For IPv6, "a Packet Too Big MUST be sent by a router in response to a packet that it cannot forward because the packet is larger than the MTU of the outgoing link"{{RFC4443}}. These PTB messages "MUST NOT be filtered" {{RFC8504}}.
- We likely do not wish to cite that - because also it does not solve the problems of PTB message delivery, but we ought to strongly encourage PTB where it can work.

gorryfair (commented):
Thanks, I expect this to be heading in a good way, and suggest more specific text to see if this helps reach final text.

draft-ietf-quic-manageability.md
chris-box: Yes, happy with that.

draft-ietf-quic-manageability.md
chris-box: Great - now applied.

draft-ietf-quic-manageability.md
chris-box: Happy with the new wording. Regarding "ought to send" my inclination is to use SHOULD but I'm not sure if normative language is permitted here. Otherwise "should".

ianswett (approved):
Thanks for the updates, this makes sense to me.

draft-ietf-quic-manageability.md
gorryfair: The document is not presently not using RFC2119 language, so "should" would seem OK personally, from my side propose what seems best.

draft-ietf-quic-manageability.md
```diff
@@ -1023,6 +1023,39 @@ not arrive at their destination. Admitting a few packets allows the QUIC
...
+lower layer before traversing a smaller MTU segment, and then reassembling.
+This is permissible even when the IP layer is IPv6 or IPv4 with the DF bit set,
+because it occurs below the IP layer. However, this process can add to compute
+and memory costs, leading to a bottleneck that limits network capacity. In such
+networks this generates a desire to influence a majority of senders to use
+smaller packets, so that the limited reassembly capacity is not exceeded.
+
+For TCP, MSS clamping (Section 3.2 of {{?RFC4459}}) is often used to change
+the sender's maximum TCP segment size, but QUIC requires a different approach.
+Section 14 of {{QUIC-TRANSPORT}} advises senders to probe larger sizes using
```
martinthomson, line 1038:
```suggestion
{{Section 14 of QUIC-TRANSPORT}} advises senders to probe larger sizes using
```

draft-ietf-quic-manageability.md
```diff
@@ -1023,6 +1023,39 @@ not arrive at their destination. Admitting a few packets allows the QUIC
...
+Datagram Packetization Layer PMTU Discovery ({{?DPLPMTUD=RFC8899}}) or Path
+Maximum Transmission Unit Discovery (PMTUD: {{?RFC1191}} and {{?RFC8201}}).
+This mechanism will encourage senders to approach the maximum size, driving
+fragmentation that they may not be aware of.
```
martinthomson, line 1042:
```suggestion
This mechanism will encourage senders to approach the maximum size, which
could cause fragmentation that they may not be aware of.
```

draft-ietf-quic-manageability.md
```diff
@@ -1023,6 +1023,39 @@ not arrive at their destination. Admitting a few packets allows the QUIC
...
+genuinely QUIC, the packet should have been a probe and so its loss should not
+cause impact to the flow of data. It serves only to guide the discovery of an
+optimal Path MTU.
+
+DPLPMTUD can optionally make use of ICMP Packet Too Big messages (PTB, defined
+in {{Section 1.1 of DPLPMTUD}}). PMTUD requires such PTB messages in order to
+work. Since the network cannot know in advance which discovery method the QUIC
+endpoints are using, it should send a PTB packet in addition to dropping the
+oversized probe packet. The generated PTB packet should be compliant with the
+validation requirements of section 14.2.1 of {{QUIC-TRANSPORT}} otherwise it
```
martinthomson, line 1056:
```suggestion
validation requirements of {{Section 14.2.1 of QUIC-TRANSPORT}} otherwise it
```

martinthomson (approved):
This is now a very good addition to the document.  Thanks Chris.

draft-ietf-quic-manageability.md
chris-box: Applied this in the latest commit.

draft-ietf-quic-manageability.md
chris-box: Why does this format give me the error "Error: IDREF attribute target references an unknown ID "section_14.2.1_of_QUIC-TRANSPORT", at None"?

draft-ietf-quic-manageability.md
```diff
@@ -1023,6 +1023,53 @@ not arrive at their destination. Admitting a few packets allows the QUIC
...
+
+If path performance is limited when sending larger packets, an on-path
+device should support a maximum packet size for a specific transport flow
+and then consistently drop all packets that exceed the configured size
+when the inner IPv4 packet has DF set, or IPv6 is used. Endpoints can cache
+PMTU information between IP flows, in the IP-layer cache, so short-term
+consistency between the PMTU for flows can help avoid an endpoint using a
+PMTU that is inefficient.
+
+If network operators wish to start with a more limited implementation of
```
gorryfair, line 1052: /more limited implementation of ... packet drops/  does read rather odd, I think this needs to say /drops to avoid fragmentation/. e.g.;
"if network operators wish to start to drop large packets to avoid fragmentation,

draft-ietf-quic-manageability.md
```diff
@@ -1023,6 +1023,53 @@ not arrive at their destination. Admitting a few packets allows the QUIC
...
+networks this generates a desire to influence a majority of senders to use
+smaller packets, so that the limited reassembly capacity is not exceeded.
+
+For TCP, MSS clamping (Section 3.2 of {{?RFC4459}}) is often used to change
+the sender's maximum TCP segment size, but QUIC requires a different approach.
+{{Section 14 of QUIC-TRANSPORT}} advises senders to probe larger sizes using
+Datagram Packetization Layer PMTU Discovery ({{?DPLPMTUD=RFC8899}}) or Path
+Maximum Transmission Unit Discovery (PMTUD: {{?RFC1191}} and {{?RFC8201}}).
+This mechanism will encourage senders to approach the maximum size, which
+could cause fragmentation that they may not be aware of.
```
gorryfair, line 1042: could this be /fragmentation with a network segment/ ... just to be clear that this isn't something else?

gorryfair (commented):
I think this is good to commit, with two minor issues.

draft-ietf-quic-manageability.md
martinthomson: Try updating kramdown-rfc2629; it's a recent feature.

chris-box: I was using version 1.3.35; updating to 1.3.37 didn't help. However in debugging I realised that it's a case sensitivity problem. When the feature was introduced in 1.3.21 the notes said "Limitations for now: Only recognizes Section/Sections".

This means you cannot write {{section 14.2.1 of RFC1149}} but must write {{Section 14.2.1 of RFC1149}}. I know that was exactly the change you suggested but the resulting I-D then has an incongruous capital S in the middle of the sentence: "validation requirements of Section 14.2.1 of ". The output format is a feature of xml2rfc so I guess I have to accept that it must always be written as Section rather than section.

draft-ietf-quic-manageability.md
chris-box: Done

draft-ietf-quic-manageability.md
chris-box: Done

draft-ietf-quic-manageability.md
martinthomson: It's not Carsten being obstinately German, it's good writing apparently.  Section is always capitalized as a proper noun.  It might not being especially proper, but that's the convention I was taught.

draft-ietf-quic-manageability.md
```diff
@@ -1023,6 +1025,53 @@ not arrive at their destination. Admitting a few packets allows the QUIC
...
+consistency between the PMTU for flows can help avoid an endpoint using a
+PMTU that is inefficient.
+
+If network operators wish to start to drop large packets to avoid
+fragmentation, they may wish to focus on QUIC. QUIC packets cannot always
+be identified ({{sec-identifying}}), but we assume that a set of candidate
+QUIC flows have been chosen. For QUIC endpoints using DPLPMTUD it is
+minimally sufficient for the path to silently drop a packet larger than the
+supported size. (A QUIC probe packet is used to discover the PMTU. If lost,
+this does not impact the flow of QUIC data.)
```
mirjak, line 1060:
```suggestion
If network operators wish to drop large packets to avoid
fragmentation, they may wish to focus on QUIC. QUIC packets cannot always
be identified ({{sec-identifying}}), but we assume that a set of candidate
QUIC flows have been chosen. For QUIC endpoints using DPLPMTUD it is
minimally sufficient for the path to silently drop a packet larger than the
supported size. (A QUIC probe packet is used to discover the PMTU. If lost,
this does not impact the flow of QUIC data.)
```

mirjak, line 1060:
```suggestion
Dropping QUIC packets that are used by DPLPMTUD to probe the path MTU will
not impact QUIC performance but can prevent a QUIC flow from increasing its packet
size.
Even though QUIC packets cannot be unambiguously identified ({{sec-identifying}} as
probes, network with configurations that would lead to fragmentation of large packets 
should rather drop large QUIC packet than fragmenting them as this will likely provide
the right signal to the endpoint to keep the packet size small and thereby avoid network
fragmentation for that flow entirely.
```

chris-box: Mirja I can see the value in amending the wording from "does not impact the flow of QUIC data" to "will not impact QUIC performance but can prevent a QUIC flow from increasing its packet size". Happy with that. But what your suggested change removes are a couple of notions. One is that it's a valid/common operator choice to be cautious in deploying packet drops, by choosing to focus on something well-known (QUIC, in this case) rather than all large IP. Secondly, the notion that QUIC itself can be hard to identify. It's not always as obvious as UDP/443. You refer instead to the ability to identify probe packets from within a known QUIC flow. Are you explicitly wanting to remove both of these, or are these merely side-effects of your suggested change?

mirjak: I wasn't trying to change the content here but to make it more clear what the recommendation is. I was working on the slides for the meeting and was looking for one sentence to copy in the slides that actually makes a recommendation (like my second proposed sentence now) but could find it.

I didn't realise that you are also talking about identifying QUIC (with respect to other UDP traffic) as you talk about QUIC packets and not QUIC flows in the original text but I left the reference to sec-identifying which should give the reader enough context anyway.

The recommendation explicitly says "QUIC packets" so I thought it was sufficiently clear that we basically say "yes we understand that you want to careful about dropping packets but if for QUIC is would be really good to drop anyway". I guess we could add back another sentence to make that bit more clear. Again my intention really was to make more clear what the recommendation actually is; so was looking for a sentence with "should" in it ;-)

chris-box: You'll find two sentences with recommendations ("should") at the end.

As for this particular paragraph, I propose the following merged form.

> Networks with configurations that would lead to fragmentation of large packets should drop such packets rather than fragmenting them. If network operators wish to start implementing this, they may wish to focus on QUIC. QUIC flows cannot always be identified ({{sec-identifying}}), but we assume that a set of candidate flows have been chosen. For QUIC endpoints using DPLPMTUD it is minimally sufficient for the path to silently drop a packet larger than the supported size. (A QUIC probe packet is used to discover the PMTU. If lost, this does not impact the flow of QUIC data.)

I'm trying to avoid saying networks should _just_ drop these packets. The recommendation is to drop and send PTB. But sometimes PTB won't be possible.

So I also propose using your sentence in the final paragraph:

> Since a network cannot know in advance which discovery method a QUIC endpoint is using, it should always send a PTB message in addition to dropping the oversized packet. A generated PTB message should be compliant with the validation requirements of {{Section 14.2.1 of QUIC-TRANSPORT}}, otherwise it will be ignored by DPLPMTUD. This will likely provide the right signal for the endpoint to keep the packet size small and thereby avoid network fragmentation for
> that flow entirely.

chris-box: Happy to do nearly all the above. The only part I don't think is needed are "based on e.g. handshake semantics or the connection ID". I'll add a commit after the meeting.

mirjak: Okay. Thx!

PR #213, opened 2021-02-08
reword "interference by firewalls"

### LINKED ISSUE #196: App: ... interference by network elements such as firewalls
The ID said: “blocking or other interference by network elements such as firewalls”
- I think this language could be seen as having a little attitude that could be avoided. I suggest replacing “interference”, and instead say that this can change the forwarding policy. This also extends it to include action from QoS policies and service level agreements.
“by network elements such as firewalls that rely on the port number for application identification.”
- I’d prefer “use” to “rely” … because they might well use other heuristics in future if they do prove more useful to classify sessions/packets.

mirjak:
fixes #196

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -478,8 +478,7 @@ practice that a port number maps unambiguously to an application, the
 use of ports that cannot easily be mapped to a registered service name
 might lead to blocking or other changes to the forwarding behavior by network
 elements such as firewalls that use the port number for application
-identification, often based on QoS policies or service level agreements
-(SLAs).
+identification.
 
 
 # Connection Migration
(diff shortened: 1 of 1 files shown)


gorryfair:
The proposed text looks OK to me.

draft-ietf-quic-applicability.md
```diff
@@ -476,8 +476,10 @@ between the endpoints on the protocol stack in use.
 Given the prevalence of the assumption in network management
 practice that a port number maps unambiguously to an application, the
 use of ports that cannot easily be mapped to a registered service name
-might lead to blocking or other interference by network elements such as
-firewalls that rely on the port number for application identification.
+might lead to blocking or other changes to the forwarding behavior by network
+elements such as firewalls that use the port number for application
+identification, often based on QoS policies or service level agreements
+(SLAs).
```
martinthomson, line 482: Not sure that I understand the rationale for the last clause here.  "often" implies that you are making a claim, but I don't believe that this one can be substantiated.  Can that be dropped?

```suggestion
identification.
```

mirjak: @gorryfair proposed to add this. However, I also don't think this adds much (or might be even a bit confusing).

gorryfair: The new text looks good. Martin's edit is also fine with me. ((I prefer  we omit the examples of other ways the forwarding treatment can be changed - these are not needed.))

PR #214, opened 2021-02-08
Clarify routing stability requirement

### LINKED ISSUE #206: App/Ops: Server deployments with load balancers
“Server deployments with load balancers and other routing infrastructure need to ensure that this infrastructure consistently routes packets to the correct server instance. “ could this be:
“Server deployments with load balancers and other routing infrastructure that depends on observing the payload of IP packets need to ensure that this infrastructure consistently routes all the QUIC packet belonging to a connection to the same server instance.”

- Seems like that is helpful. Is there a relevant section in the OPS draft where the reader should be directed?

martinduke:
Fixes #206. 

Does this make it easier to understand?


mirjak:
@gorryfair is this better?

gorryfair (approved):
looks good to me.

PR #220, opened 2021-02-12
clarify that ports are advisory for the network

### LINKED ISSUE #176: Section 2.3 - QUIC Port Numbers (not really different to other ports)
It seems to me that QUIC is no different to other transports in respect of being able to utilise other port numbers, and it would be worth noting that this is normal. The present text might perhaps suggest this is specific to QUIC. Should the text reference RFC 7605 or something else just to explain ports don't imply specific application behaviour, but could infer this?.

britram:
fix #176

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -224,11 +224,11 @@ ID; see Section 17.2 of {{QUIC-TRANSPORT}}.
 ## Use of Port Numbers
 
 Applications that have a mapping for TCP as well as QUIC are expected to
-use the same port number for both services. However, as with
-TCP-based services, especially when application layer information is
-encrypted, there is no guarantee that a specific application will use the
-registered port, or the used port is carrying traffic belonging to the
-respective registered service. For example, {{QUIC-HTTP}} specifies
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
The title of the pull request was cause for a momentary flutter - in that I don't see the port as carrying anything other than a reference to a convention - but the text is very good.

draft-ietf-quic-manageability.md
```diff
@@ -224,11 +224,11 @@ ID; see Section 17.2 of {{QUIC-TRANSPORT}}.
 ## Use of Port Numbers
 
 Applications that have a mapping for TCP as well as QUIC are expected to
-use the same port number for both services. However, as with
-TCP-based services, especially when application layer information is
-encrypted, there is no guarantee that a specific application will use the
-registered port, or the used port is carrying traffic belonging to the
-respective registered service. For example, {{QUIC-HTTP}} specifies
+use the same port number for both services. However, as all other IETF
```
mirjak, line 227:
```suggestion
use the same port number for both services. However, as for all other IETF
```

mirjak (approved):

PR #222, opened 2021-02-12
UDP Timeout text

### LINKED ISSUE #183: Sect 4.1 : what is the recommendation for middleboxes/NATs?
Sect 4.1 : This contains a statement on the recommendations for middleboxes and NATs, that I can’t understand, and this concerns me.
i.e. to me, the present text suggests that the advice in BCP 127 (aka RFC4787) should be changed for QUIC. I don’t recall this being discussed - was it?

I don’t actually follow the rationale that says this follows from QUIC being connection-oriented. If the connection close was visible to the device, and it could release state - then this might make sense, but that isn’t exactly what is explained. I’m left unsure how a “QUIC-aware” middlebox is supposed to track this state.

I'm also unsure how say a NAPT could know the traffic was QUIC, is this suggesting a port-based change. I think I need to understand this proposal.

### LINKED ISSUE #184: Sect 4.8: Reducing the timeout on UDP NATs
Sect 4.8: Reducing the timeout on UDP NATs

“Reducing the timeout on UDP NATs might be tempting in light of this property, but not all QUIC server deployments will be robust to rebinding.” 
- so  now,  I did not understand what was being discouraged here. Is this somehow linked to section 4.6’s discussion of holding state?

(perhaps related to issue #183)

mdukef5:
Fixes #183. Fixes #184.

I don't have a strong opinion about the point of view in this PR, though I believe it accurately reflects the discussion in the issues so far.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -790,15 +790,21 @@ signal ({{sec-teardown}}) means that this state must be purged either through
 timers or through least-recently-used eviction, depending on application
 requirements.
 
-{{?RFC4787}} requires a timout that is not less than 2 minutes for most UDP
-traffic.
-However, in pratice, timers are often lower, in the range of 15 to 30 seconds.
-In contrast, {{?RFC5382}} recommends a timeout of more than 2 hours for TCP,
-given that TCP is a connection-oriented protocol with well-defined closure
-semantics. For network devices that are QUIC-aware, it is recommended to also
-use longer timeouts for QUIC traffic, as QUIC is connection-oriented. As such,
-a handshake packet from the server indicates the willingness of the server to
-communicate with the client.
+{{?RFC4787}} requires a timeout that is not less than 2 minutes for most UDP
+traffic.  However, in pratice, timers are often lower, in the range of 15 to 30
+seconds. In contrast, {{?RFC5382}} recommends a timeout of more than 2 hours for
+TCP, given that TCP is a connection-oriented protocol with well-defined closure
+semantics.
+
+As QUIC has features to improve robustness to NAT rebinding, and has no
+observable end-of-connection semantics, devices might choose to further reduce
+the UDP timeout to avoid unnecessary state. However, this has two drawbacks.
+First, not all servers inhabit routing architectures where connections will
+survive a port or address change. Second, an intermediary might require one
+endpoint to re-instantiate the connection for security reasons, further
+reducing the likelihood of successful revival. For these reasons, devices that
+can distinguish QUIC traffic from other UDP payloads should retain the
+{{?RFC4787}} limits.
 
 The QUIC header optionally contains a connection ID which can be used as
 additional entropy beyond the 5-tuple, if needed. The QUIC handshake needs
@@ -807,7 +813,7 @@ what length it has. However, connection IDs may be renegotiated during
 a connection, and this renegotiation is not visible to the path. Keying state
 off the connection ID may therefore cause undetectable and unrecoverable loss
 of state in the middle of a connection. Use of connection ID specifically
-discouraged for NAT applications.
+discouraged for NAT applications (see {{quic-nat}}).
 
 ## Passive Network Performance Measurement and Troubleshooting
 
@@ -914,7 +920,7 @@ more tolerant of packet re-ordering than traditional TCP traffic (see
 recovery mechanism is used and therefore reordering tolerance should be
 considered as unknown.
 
-## QUIC and Network Address Translation (NAT)
+## QUIC and Network Address Translation (NAT) {#quic-nat}
 
 QUIC Connection IDs are opaque byte fields that are expressed consistently
 across all QUIC versions {{QUIC-INVARIANTS}}, see {{rebinding}}. This feature
@@ -934,14 +940,11 @@ connection-oriented protocols, why NAT use of Connection ID might appear
 attractive, and how NAT use of CID can create serious problems for the
 endpoints.
 
+See {{sec-stateful}} for a discussion of the proper timeout for NAT connection
+state.
+
 {{?RFC4787}} contains some guidance on building NATs to interact constructively
 with a wide range of applications. This section extends the discussion to QUIC.
-
-By using the CID, QUIC connections can survive NAT rebindings as long as no
-routing function in the path is dependent on client IP address and port to
-deliver packets between server and NAT. Reducing the timeout on UDP NATs might
-be tempting in light of this property, but not all QUIC server deployments will
-be robust to rebinding.
 
 ### Resource Conservation
 


ianswett (changes requested):
Let's discuss more on the issue or on VC.

draft-ietf-quic-manageability.md
```diff
@@ -790,15 +790,21 @@ signal ({{sec-teardown}}) means that this state must be purged either through
...
-traffic.
-However, in pratice, timers are often lower, in the range of 15 to 30 seconds.
-In contrast, {{?RFC5382}} recommends a timeout of more than 2 hours for TCP,
-given that TCP is a connection-oriented protocol with well-defined closure
-semantics. For network devices that are QUIC-aware, it is recommended to also
-use longer timeouts for QUIC traffic, as QUIC is connection-oriented. As such,
-a handshake packet from the server indicates the willingness of the server to
-communicate with the client.
+{{?RFC4787}} requires a timeout that is not less than 2 minutes for most UDP
+traffic.  However, in pratice, timers are often lower, in the range of 15 to 30
```
ianswett, line 794: 15 seconds is very rare unless something has changed recently, so I'd say 30 seconds to a minute.

draft-ietf-quic-manageability.md
```diff
@@ -790,15 +790,21 @@ signal ({{sec-teardown}}) means that this state must be purged either through
...
+TCP, given that TCP is a connection-oriented protocol with well-defined closure
+semantics.
+
+As QUIC has features to improve robustness to NAT rebinding, and has no
+observable end-of-connection semantics, devices might choose to further reduce
+the UDP timeout to avoid unnecessary state. However, this has two drawbacks.
+First, not all servers inhabit routing architectures where connections will
+survive a port or address change. Second, an intermediary might require one
+endpoint to re-instantiate the connection for security reasons, further
+reducing the likelihood of successful revival. For these reasons, devices that
```
ianswett, line 805: I don't think I understand this second point about re-instantiating the connection for security reason.

draft-ietf-quic-manageability.md
```diff
@@ -790,15 +790,21 @@ signal ({{sec-teardown}}) means that this state must be purged either through
...
-a handshake packet from the server indicates the willingness of the server to
-communicate with the client.
+{{?RFC4787}} requires a timeout that is not less than 2 minutes for most UDP
+traffic.  However, in pratice, timers are often lower, in the range of 15 to 30
+seconds. In contrast, {{?RFC5382}} recommends a timeout of more than 2 hours for
+TCP, given that TCP is a connection-oriented protocol with well-defined closure
+semantics.
+
+As QUIC has features to improve robustness to NAT rebinding, and has no
+observable end-of-connection semantics, devices might choose to further reduce
```
ianswett, line 800: Most UDP protocols don't have end-of-connection semantics, so this comparison seems odd in this context, since the UDP timeout is already much less than TCP.

draft-ietf-quic-manageability.md
martinduke: OK, I just took that from the original text. Happy to change it.

draft-ietf-quic-manageability.md
martinduke: This is the thing where a server packet does no good for rebinding because the firewall expects the first packet from the client. Can you suggest alternate text?

draft-ietf-quic-manageability.md
ianswett: Understood, but yes, let's change it.

draft-ietf-quic-manageability.md
ianswett, line 805:
```suggestion
survive a port or address change. Second, applications which allow either peer to send after a
period of quiescence will experience black-holing until the peer re-initializes the NAT binding.
For these reasons, devices that
```

ianswett: Suggestion below.

draft-ietf-quic-manageability.md
ianswett:
```suggestion
observable end-of-connection semantics, devices might be tempted to further reduce
```

draft-ietf-quic-manageability.md
mirjak: I thought we took this from some paper but can't real find it anymore. I found this one though which actually indicates that most timeout are 2 minutes: https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.739.5004&rep=rep1&type=pdf

draft-ietf-quic-manageability.md
```diff
@@ -790,15 +790,21 @@ signal ({{sec-teardown}}) means that this state must be purged either through
...
+the UDP timeout to avoid unnecessary state. However, this has two drawbacks.
```
ianswett, line 801:
```suggestion
As QUIC has features to improve robustness to NAT rebinding, devices might be
tempted to further reduce the UDP timeout to avoid unnecessary state. However,
this has two drawbacks.
```

I don't like 'tempted', but the current text read to me as though you're saying it's ok, which I do not want to convey.

ianswett (commented):
Some suggestions

draft-ietf-quic-manageability.md
mirjak: I'd recommend to rather start this paragraph with saying what we recommend and not saying what we no recommend :-)

draft-ietf-quic-manageability.md
```diff
@@ -790,15 +790,21 @@ signal ({{sec-teardown}}) means that this state must be purged either through
...
+First, not all servers inhabit routing architectures where connections will
+survive a port or address change. Second, an intermediary might require one
+endpoint to re-instantiate the connection for security reasons, further
+reducing the likelihood of successful revival. For these reasons, devices that
+can distinguish QUIC traffic from other UDP payloads should retain the
+{{?RFC4787}} limits.
```
mirjak, line 807: I know the originally text wasn't that much better, but I guess actually all devices should retain the RFC4787 limits for all UDP traffic because that's what we recommend in RFC4787. This text should only say that it even more important and also more reasonable to retain these limits for QUIC because QUIC is connection-oriented and therefore a device can at least because sure that there is a valid communication going on if it sees an successful handshake. 

I actually don't think, it a good advise to tell these boxes to be QUIC aware. It's only if you really don't want to adjust your timers for other UDP traffic than you should still consider doing so for QUIC specifically which however means you need additional logic to detect QUIC traffic, which however in itself is not super easy (see other section).

draft-ietf-quic-manageability.md
martinduke: accepted this manually

martinduke:
Alright, I believe I've addressed all the concerns, assuming we want to move forward with a "don't change the timeout" recommendation. PTAL.

draft-ietf-quic-manageability.md
```diff
@@ -790,15 +790,21 @@ signal ({{sec-teardown}}) means that this state must be purged either through
...
-However, in pratice, timers are often lower, in the range of 15 to 30 seconds.
-In contrast, {{?RFC5382}} recommends a timeout of more than 2 hours for TCP,
-given that TCP is a connection-oriented protocol with well-defined closure
-semantics. For network devices that are QUIC-aware, it is recommended to also
-use longer timeouts for QUIC traffic, as QUIC is connection-oriented. As such,
-a handshake packet from the server indicates the willingness of the server to
-communicate with the client.
+{{?RFC4787}} requires a timeout that is not less than 2 minutes for most UDP
+traffic.  However, in pratice, timers are sometimes lower, in the range of 30 to
+60 seconds. In contrast, {{?RFC5382}} recommends a timeout of more than 2 hours
```
martinthomson, line 795:
```suggestion
traffic.  However, in practice, timers are sometimes lower, in the range of 30 to
60 seconds. In contrast, {{?RFC5382}} recommends a timeout of more than 2 hours
```

draft-ietf-quic-manageability.md
```diff
@@ -790,15 +790,21 @@ signal ({{sec-teardown}}) means that this state must be purged either through
...
+for TCP, given that TCP is a connection-oriented protocol with well-defined
+closure semantics.
+
+As QUIC has features to improve robustness to NAT rebinding, devices might be
+tempted to further reduce the UDP timeout to avoid unnecessary state. However,
+this has two drawbacks.
+
+First, not all servers inhabit routing architectures where connections will
```
martinthomson, line 803:
```suggestion
First, not all endpoints use routing architectures where connections will
```

Not sure about endpoints vs. servers here, but "inhabit" implies life.

draft-ietf-quic-manageability.md
```diff
@@ -790,15 +790,21 @@ signal ({{sec-teardown}}) means that this state must be purged either through
...
+survive a port or address change. Second, applications which allow either peer
+to send after a period of quiescence will experience black-holing until the peer
+re-initializes the NAT binding. For these reasons, devices that can distinguish
+QUIC traffic from other UDP payloads should retain the {{?RFC4787}} limits.
```
martinthomson, line 807:
```suggestion
re-initializes the NAT binding. For these reasons, retain the {{?RFC4787}} limits is
useful, even where devices are able to distinguish QUIC traffic from other UDP payloads.
```

martinthomson (approved):

draft-ietf-quic-manageability.md
```diff
@@ -783,22 +783,28 @@ techniques and how QUIC's design impacts them.
...
+As QUIC has features to improve robustness to NAT rebinding, devices might be
+tempted to further reduce the UDP timeout to avoid unnecessary state. However,
+this has two drawbacks.
+
+First, not all endpoints use routing architectures where connections will
+survive a port or address change. Second, applications which allow either peer
+to send after a period of quiescence will experience black-holing until the peer
+re-initializes the NAT binding. For these reasons, retaining the {{?RFC4787}}
+limits is useful, even where devices are able to distinguish QUIC traffic from
+other UDP payloads.
```
mirjak, line 807: Not sure if the advise is crystal clear here now. Initially the advise was more like "even if you use too short time-outs for UDP, please please don't do that for QUIC and use at least two minutes". While now the recommendation here rather reads like "just do the same thing as you do for UDP".

gorryfair: I think I agree, so can we change this: "For these reasons, retaining the {{?RFC4787}} limits is useful, even where devices are able to distinguish QUIC traffic from other UDP payloads." to something stronger, e.g. perhaps like:
"For these reasons, the limits in {{?RFC4787}} are important to avoid black-holing of packets (and hence avoid interrupting the flow of data to the client), even where devices are able to distinguish QUIC traffic from other UDP payloads."

gorryfair (commented):
Nearly there... I think this needs slightly more encouragement to set the timers correctly.

draft-ietf-quic-manageability.md
mirjak, line 807:
```suggestion
Even though QUIC has explicitly been designed to improve robustness to NAT rebindings,
which might look like a tempting feature to further reduce the UDP timeout and thereby
minimize state, this is not recommended.  Instead it is recommended, even when lower
timers are used for other UDP traffic, to use a timer of at least two minutes for QUIC traffic.

While QUIC has no clear network-visible end-of-connection signal and therefore
does require timer-based state removal, the QUIC handshake indicates confirmation
of both ends that a valid bidirectional transmission is on-going. If state is removed
too early, this could lead to black-holing of incoming packets after a short idle period.
To detect this situation, a timer at the client needs to expire before a re-establishment
can happen (if at all), which would lead to unnecessary long delays in an otherwise
working connection.

Furrther, not all endpoints use routing architectures where connections will
survive a port or address change. So even when the client revives the connection,
a NAT rebinding can cause a routing mismatch where a packet is not even delivered to
the server that might support address migration. 

For these reasons, the limits in {{?RFC4787}} are important to avoid black-holing
of packets (and hence avoid interrupting the flow of data to the client), especially where
devices are able to distinguish QUIC traffic from other UDP payloads.
```

Just wanted to make a proposal to make the recommendation a bit stronger but then ended up with quite a bit a rewrite... so here is my proposal...

gorryfair: Looks better to me!

Check: /Furrther,/Further,/ or Furthermore,

ianswett:
Q: Is this still an active PR?

mirjak:
this PR should be replaced with #300 (where just waiting for some feedback from @martinduke before closing...)

britram:
superceded by #300

PR #223, opened 2021-02-12
Editorial nits from @gorryfair, thanks!

### LINKED ISSUE #185: This issue contains a set of editorial NiTs from QUIC Manageability-09
This section contains a series, of assumed editorial comments. (The issue is copied from the QUIC WG list).
“Even thought this bit is fixed in the QUICv1 specification,”
-  This seems like the wrong word: /thought/though/
---
“observers cannot reliably use“
-Maybe we could avoid “reliably”, e.g.:
 /cannot depend on this bit as/...
---
“first octet in the short packet header.”
would probably benefit from an indication of bit ordering and cross-reference to the QUIC spec.
---
Sect 2.4: “In the nominal case” - this use of “nominal” was unusual to me, the word is used differently in different English-speaking regions - in some it might mean “marginal” or only correct at the current time - in other regions it might mean something else. Is “normal” acceptable in this sentence?
---
Section 2.3: “We therefore consider” - who is the “we”, suggest something like “These are considered...”
---
“in the clear” - would be simpler as “without encryption” or “observable on-path”?
---
The term “exposes” is used several times, without saying to whom, which I think in the context of this document, could be clearer as “exposes this, making it visible to an on-path observer”. Also this form of wording is consistent with sect 2.7. This also could help explain this in a more positive light to someone wishing to manage a network.
---
Sect 2.4: “The amount of 0-RTT protected data is limited by the initial congestion window, typically around 10 packets [RFC6928].”
- I suggest this should instead be reference to a section in the QUIC spec. (This does in turn point to 6928, but I think the dependency here is on what QUIC says.)
---
Sect 2.8”: ...”any manipulation of ... will be detected”
I think the word “manipulation” hints like it is illicit, can we use neutral language by saying something like “any change by network devices on-path”?
---
Sect 3.1: “First, it only provides one bit of information and is quite prone to collide with UDP-based protocols other than those that this static bit is meant to allow multiplexing with.“
This seems to lack a little polish (“quite prone”, “meant to ...with”).
---
Sect 3.3.1.
/through passive measurement/by observation by an on-path observer/
... I think this applies, whether it is measurement, filtering, whatever, so we could use the same term.
---
/which each one byte of zeros/ ... seems an odd set of words to me.
---
Section 4.1: “Keying state off the connection ID may therefore cause undetectable and unrecoverable loss of state in the middle of a connection.  Use of connection ID specifically discouraged for NAT applications.”
... I didn’t find /Keying state/ very helpful... maybe some other words would be clearer. The second sentence seems to miss /is/?
---
Sect 4.6: “ Distinguishing ACK packets is trivial in TCP,“
- well, this is not really trivial, because there are pure ACKs and data+ACK which makes this complicated. RFC3449 says more.
---

britram:
Fixes #185.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -532,7 +532,7 @@ see also {{?RFC7605}} section 5.
 While the second most significant bit (0x40) of the first octet is set to
 1 in most QUIC packets of the current version (see {{public-header}} and
 section 17 of {{QUIC-TRANSPORT}}), this method of recognizing QUIC traffic
-is NOT RECOMMENDED. First, it only provides one bit of information and is
+is not reliable. First, it only provides one bit of information and is
 prone to collision with UDP-based protocols other than those that this static
 bit is meant to allow multiplexing with. Second, this feature of the wire
 image is not invariant {{QUIC-INVARIANTS}} and may change in future versions
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -805,9 +805,9 @@ The QUIC header optionally contains a connection ID which can be used as
 additional entropy beyond the 5-tuple, if needed. The QUIC handshake needs
 to be observed in order to understand whether the connection ID is present and
 what length it has. However, connection IDs may be renegotiated during
-after the handshake, and this renegotiation is not visible to the path. Using the
-connection ID as a flow key field for stateful treatment of flows may therefore
-cause undetectable and unrecoverable loss
+after the handshake, and this renegotiation is not visible to the path. Using 
+the connection ID as a flow key field for stateful treatment of flows may 
+therefore cause undetectable and unrecoverable loss
 of state in the middle of a connection. Use of connection IDs is specifically
 discouraged for NAT applications.
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -805,7 +805,7 @@ The QUIC header optionally contains a connection ID which can be used as
 additional entropy beyond the 5-tuple, if needed. The QUIC handshake needs
 to be observed in order to understand whether the connection ID is present and
 what length it has. However, connection IDs may be renegotiated during
-a connection, and this renegotiation is not visible to the path. Using the
+after the handshake, and this renegotiation is not visible to the path. Using the
 connection ID as a flow key field for stateful treatment of flows may therefore
 cause undetectable and unrecoverable loss
 of state in the middle of a connection. Use of connection IDs is specifically
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -428,7 +428,7 @@ bytes. Additional datagrams containing only 0-RTT protected long header packets
 may be sent from the client to the server after the Client Hello datagram,
 containing the rest of the 0-RTT data. The amount of 0-RTT protected data is
 limited by the initial congestion window, typically around 10 packets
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -530,13 +530,14 @@ given flow is using QUIC based upon a UDP port number may therefore not hold;
...
-1 in most QUIC packets of the current version (see {{public-header}}),
-this method of recognizing QUIC traffic is NOT RECOMMENDED. First, it only
-provides one bit of information and is quite prone to collide with
-UDP-based protocols other than those that this static bit is meant to allow
-multiplexing with. Second, this feature of the wire image is not invariant
-{{QUIC-INVARIANTS}} and may change in future versions of the protocol, or
-even be negotiated during the handshake via the use of transport parameters.
+1 in most QUIC packets of the current version (see {{public-header}} and
+section 17 of {{QUIC-TRANSPORT}}), this method of recognizing QUIC traffic
+is NOT RECOMMENDED. First, it only provides one bit of information and is
```
mirjak, line 535:
```suggestion
is not sustainable. First, it only provides one bit of information and is
```

mirjak (approved):

draft-ietf-quic-manageability.md
ianswett:
```suggestion
is not reliable. First, it only provides one bit of information and is
```

draft-ietf-quic-manageability.md
```diff
@@ -804,8 +805,9 @@ The QUIC header optionally contains a connection ID which can be used as
 additional entropy beyond the 5-tuple, if needed. The QUIC handshake needs
 to be observed in order to understand whether the connection ID is present and
 what length it has. However, connection IDs may be renegotiated during
-a connection, and this renegotiation is not visible to the path. Keying state
-off the connection ID may therefore cause undetectable and unrecoverable loss
+a connection, and this renegotiation is not visible to the path. Using the
+connection ID as a flow key field for statful treatment of flows may therefore
```
ianswett, line 809:
```suggestion
connection ID as a flow key field for stateful treatment of flows may therefore
```

draft-ietf-quic-manageability.md
```diff
@@ -804,8 +805,9 @@ The QUIC header optionally contains a connection ID which can be used as
...
+cause undetectable and unrecoverable loss
 of state in the middle of a connection. Use of connection ID specifically
```
ianswett, line 811:
```suggestion
of state in the middle of a connection. Use of connection IDs is specifically
```

draft-ietf-quic-manageability.md
```diff
@@ -804,8 +805,9 @@ The QUIC header optionally contains a connection ID which can be used as
 additional entropy beyond the 5-tuple, if needed. The QUIC handshake needs
 to be observed in order to understand whether the connection ID is present and
 what length it has. However, connection IDs may be renegotiated during
-a connection, and this renegotiation is not visible to the path. Keying state
-off the connection ID may therefore cause undetectable and unrecoverable loss
+a connection, and this renegotiation is not visible to the path. Using the
```
ianswett, line 808:
```suggestion
after the handshake, and this renegotiation is not visible to the path. Using the
```

draft-ietf-quic-manageability.md
ianswett: This sentence isn't quite as cautionary as I think it should be, but I don't have any good suggestion.

ianswett (commented):
LG, some small suggestions.

martinthomson (approved):
...with Ian's suggestions.

Issue #197, opened 2021-02-06
App: What does "prime a CC" mean?


gorryfair:
The text says: “or prime the congestion controller in preparation for switching to the new path.” 
- what does prime mean?
-  and which section explains this function  in the QUIC spec?

mirjak:
The idea here is that you can already e.g. estimate RTT or even open the congestion window before you switch. There was already some discussion previously (seePR #122) if we want to have this in the document or say more or less. Maybe a question to the rest of the wg (in the next meeting)...?

gorryfair:
Unsure about this in this ID: Isn't that suggesting a transport extension/mechanism update... rather than the applicable use?

mirjak:
I don't think it needs an extension. It can simply be done by the client. It's definitely not appropriate to specify a concrete algorithm in this document but it's just listed as one of the things that padding can be used during mitigation.

gorryfair:
I'm really unsure about why text about priming is here and where the advice is located. I don't dispute that it can be done, or even that if done appropriately it can help, however it could also be done in a way that might also hinder. If PR 122 removes this word, then this topic can be discussed elsewhere in RFCs.

mirjak:
@gorryfair PR #122 was already merged a long time ago. I was just point at it to say that there has been previous discussion. As the label indicates, we plan to bring the for discussion to the meeting.

martinduke:
CC priming seems like an issue entirely internal to the QUIC implementation. Why should applications know or care about it?

I suggest we simply delete this.

martinthomson:
I like @martinduke's suggestion.

mirjak:
#291 is my proposal. It doesn't take about priming/pre-heating anymore but it says what's possible and why this icould be interesting for an application.

mirjak:
PR #291 was merged.

PR #122, opened 2020-08-24
Update section on path migration

mirjak:
addresses #115 (partially at least)

PR #225, opened 2021-02-12
Editorial rollup for applicability

### LINKED ISSUE #186: applicability: ID Nits: unused reference
Unused Reference: 'I-D.nottingham-httpbis-retry' is defined on line 861, but no explicit reference was found in the text.

### LINKED ISSUE #187: applicability: ID Nits: duplicate reference
Duplicate reference: RFC7301, mentioned in 'RFC7301', was also mentioned in 'ALPN'.

### LINKED ISSUE #205: App: Many and various editorial NiTS
The app ID seems to have various places where the text needs a wee polish. I'm sure this is not all, but there are quite a lot of NiTS here. It could be that some of these may be hints at topics that should become issues to be discussed, although I suspect that most (or all) are simply requests for editorial changes to improve clarity.

---
NiT: Timeliness of the “evidence”.
I agree with what is said, but I think we should be more cautious about current evidence we place in the RFC. 
“While there is no evidence of widespread, systematic disadvantage”
- can we say “Current measurements show no evidence…” or something that captures this.
---
Nit: “Robust to rebinding” might be better as “robust to NAT rebinding”.
---
Nit: “can use 0-RTT session resumption,”, I’d suggest a cross reference to section here in the QUIC Spec.
---
Nit: “connections may not be robust”, this combination of may can often be misinterpreted by different usages of English, “might not” is likely unambiguous.
---
Nit: “In particular, firewalls will often not admit server traffic for which it has not kept state for corresponding packets from the client.” I think this is about what I would call “established” state tracking - mixed plurals, could this be something a little like: “In particular, some firewalls do not admit server traffic for which the firewall has no recent state for  a corresponding packet sent from the client.”
---
Nit: “Sending PING frames more frequently than every 30 seconds over long idle periods may result in excessive unproductive traffic in some”  - could be another place to refer to RFC8085 o/andr QUIC Spec section.
---
Nit: “This of course only applies “ - is it possible to avoid “of course” here? - or if needed, then please do explain more. Maybe “This approach is valid only when…” or something similar?
---
NiT: “Using resumption in this way also assumes that the protocol does not accumulate any non-persistent state in association with a connection.” - I’m not sure what the assumption is here? Isn’t this a requirement on the application?
---
Nit: “Due to encoding limitations on “. This would read easier if it did not start with “due to”.
---
Nit: “exceptional handling “ - is this “exception handling” - or is this something else then maybe this needs discussion?
---
Nit: “If a QUIC receiver has maximum allowed concurrent streams open” - I think this should be:
“If a QUIC receiver opened has the maximum allowed concurrent streams “
---
Nit: “valuable to expose maximum number of allowed” insert “the”?
---
Nit: for the avoidance of doubt: “will try to maximally pack packets” - please add “QUIC” before packets.
---
Nit: “may be valuable to indicate to QUIC that all data should be send out immediately” … I agree, but think this means the QUIC sending stack, because I don’t know the QUIC transport protocol mechanism to do this.
---
Nit: Sect 7: “If a lower-layer address changes” - what is this? an IP address? something more?
and “can still associate packets with an existing connection” - I’d prefer “transport” or “endpoint” before connection here, because there are connections at other lower layers, which this isn’t about.
---
Nit: “Path validation means that the other endpoint in required” - /other/remote/ ?? /in/is/
---
Nit: “congestion control will also be reset on path migration” is /on/ better /after/ - please refer to the relevant section in the spec.
---
Nit: “Probing packets, which cannot carry application data, can be sent on multiple paths at once.” . I think should be “QUIC packet probes” and would be much better with a cross reference to the sections within the spec.
---
Nit: “This frames has two sets of types: one” /This/These/  and /has/have/.
---
Nit: “Aside from the destination connection ID field of the first packets sent by clients, the connection ID can be zero length. “.  “Aside from” reads a little odd, but the intent seems correct, can you reword?
---
Nit: “The most efficient mitigation for these attacks is operational, either by using a load balancing architecture” Agree, but “operational” seems like the sentence is odd, is this appropriate operational practice, or design or something else?
---

### LINKED ISSUE #216: Very long paragraph in Streams section
> Mapping of application data to streams is application-specific and described for HTTP/3 in [QUIC-HTTP]. In general, data that can be processed independently, and therefore would suffer from head of line blocking if forced to be received in order, should be transmitted over separate streams. If the application requires certain data to be received in order, that data should be sent on the same stream. If there is a logical grouping of data chunks or messages, streams can be reused, or a new stream can be opened for each chunk/message. If one message is mapped to a single stream, resetting the stream to expire an unacknowledged message can be used to emulate partial reliability on a message basis. If a QUIC receiver has maximum allowed concurrent streams open and the sender on the other end indicates that more streams are needed, it doesn't automatically lead to an increase of the maximum number of streams by the receiver. Therefore it can be valuable to expose maximum number of allowed, currently open and currently used streams to the application to make the mapping of data to streams dependent on this information.

That's a lot to work through.  Consider breaking this up a little.  I see a few things:

1. Concurrency is aided by using more streams where ordering is not necessary or where concurrency is desirable.
2. On the opposite end of things, streams provide ordering where ordering is critical.
3. Sending a message per stream allows messages to be cancelled.
4. Limits on concurrency need to be accounted for.

britram:
Fixes #205 (thanks @gorryfair!); fixes #186 and fixes #187 (thanks @seanturner!). Fixes #216 as well (thanks @martinthomson!)

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -328,7 +328,7 @@ If a QUIC receiver has opened the maximum allowed concurrent
 streams, and the sender indicates that more streams are needed, it
 doesn't automatically lead to an increase of the maximum number of streams by
 the receiver. Therefore it can be valuable to expose the maximum number of 
-allowed, currently open and currently used streams to the application to make
+allowed, currently open, and currently used streams to the application to make
 the mapping of data to streams dependent on this information.
 
 While a QUIC implementation must necessarily provide a way for an application
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -310,7 +310,7 @@ the higher layer, which can eventually reset the QUIC connection.
 
 Mapping of application data to streams is application-specific and described for
 HTTP/3 in {{QUIC-HTTP}}. There are a few general principles to apply when
-designing an application's use of streams.
+designing an application's use of streams:
 
 - A single stream provides ordering. If the application requires certain data to
 be received in order, that data should be sent on the same stream.
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -428,7 +428,7 @@ QUIC provides an interface that provides multiple streams to the application;
 however, the application usually cannot control how data transmitted over one
 stream is mapped into frames or how those frames are bundled into packets.
 
-By default, many QUIC implementations will try to maximally pack QUIC packets
+By default, many implementations will try to maximally pack QUIC packets
 with one or more stream data frames to minimize bandwidth consumption and
 computational costs (see section 13 of {{!QUIC}}). If there is not enough data
 available to fill a packet, an implementation might wait for a short time, to
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -267,7 +267,7 @@ data to a server over a connection idle longer than the server's idle timeout
 (available from the idle_timeout transport parameter) can simply reconnect. When
 possible, this reconnection can use 0-RTT session resumption, reducing the
 latency involved with restarting the connection. This approach is only valid in
-cases in which 0-RTT data is safe, when the client is the restarting peer,
+cases in which 0-RTT data is safe, when the client is the restarting peer, and
 when the data to be sent is idempotent. It is also not applicable when the
 application binds external state to the connection, as this state cannot
 reliably be transferred to a resumed connection.
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -83,7 +83,6 @@ informative:
         ins: P. Sarolahti
       -
         ins: M. Kojo
-  I-D.nottingham-httpbis-retry:
 --- abstract
 
 This document discusses the applicability of the QUIC transport protocol,
@@ -117,19 +116,19 @@ transport for their application.
 QUIC uses UDP as a substrate. This enables both userspace implementation
 traversal of middleboxes and NAT without requiring updates.
 
-While there is no evidence of widespread, systematic disadvantage of UDP
-traffic compared to TCP in the Internet {{Edeline16}}, somewhere between three
-{{Trammell16}} and five {{Swett16}} percent of networks simply block UDP
-traffic. All applications running on top of QUIC must therefore either be
-prepared to accept connectivity failure on such networks, or be engineered to
-fall back to some other transport protocol. In the case of HTTP, this fallback
-is TLS 1.3 over TCP.
+While recent measurements have shown no evidence of widespread, systematic
+disadvantage of UDP traffic compared to TCP in the Internet {{Edeline16}},
+somewhere between three {{Trammell16}} and five {{Swett16}} percent of networks
+simply block UDP traffic. All applications running on top of QUIC must 
+therefore either be prepared to accept connectivity failure on such networks,
+or be engineered to fall back to some other transport protocol. In the case of
+HTTP, this fallback is TLS 1.3 over TCP.
 
 An application that implements fallback needs to consider the security
 consequences. A fallback to TCP and TLS 1.3 exposes control information to
 modification and manipulation in the network. Further downgrades to older TLS
 versions might result in significantly weaker cryptographic protection. For
-example, the results of protocol negotiation {{?ALPN=RFC7301}} only have
+example, the results of protocol negotiation {{RFC7301}} only have
 confidentiality protection if TLS 1.3 is used.
 
 These applications must operate, perhaps with impaired functionality, in the
@@ -223,12 +222,12 @@ rebinding after a timeout. However, this only helps if one endpoint maintains
 availability at the address its peer uses, and the peer is the one to send
 after the timeout occurs.
 
-Some QUIC connections may not be robust to rebinding because the routing
+Some QUIC connections might not be robust to NAT rebinding because the routing
 infrastructure (in particular, load balancers) uses the address/port four-tuple
 to direct traffic. Furthermore, middleboxes with functions other than address
-translation could still affect the path. In particular, firewalls will often
-not admit server traffic for which it has not kept state for corresponding
-packets from the client.
+translation could still affect the path. In particular, some firewalls do not
+admit server traffic for which the firewall has no recent state for a
+corresponding packet sent from the client.
 
 A QUIC application can adjust idle periods to manage the risk of timeout
 (noting that idle periods and the network idle timeout is distinct from the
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-applicability.md
```diff
@@ -260,18 +259,18 @@ periods may result in excessive unproductive traffic in some situations, and to
...
 (available from the idle_timeout transport parameter) can simply reconnect. When
 possible, this reconnection can use 0-RTT session resumption, reducing the
-latency involved with restarting the connection. This of course only applies in
-cases in which 0-RTT data is safe, when the client is the restarting peer, and
-when the data to be sent is idempotent.  Using resumption in this way also
-assumes that the protocol does not accumulate any non-persistent state in
-association with a connection.  State bound to a connection cannot reliably be
-transferred to a resumed connection.
+latency involved with restarting the connection. This approach is only valid in
+cases in which 0-RTT data is safe, when the client is the restarting peer,
```
martinthomson, line 270: Missing and

draft-ietf-quic-applicability.md
```diff
@@ -309,20 +309,27 @@ generate error messages on the application layer to inform the other end and/or
 the higher layer, which can eventually reset the QUIC connection.
 
 Mapping of application data to streams is application-specific and described for
-HTTP/3 in {{QUIC-HTTP}}. In general, data that can be processed independently,
+HTTP/3 in {{QUIC-HTTP}}. There are a few general principles to apply when
+designing an application's use of streams.
```
martinthomson, line 313: Colon

draft-ietf-quic-applicability.md
```diff
@@ -309,20 +309,27 @@ generate error messages on the application layer to inform the other end and/or
...
+on a message basis. 
+
+If a QUIC receiver has opened the maximum allowed concurrent
+streams, and the sender indicates that more streams are needed, it
 doesn't automatically lead to an increase of the maximum number of streams by
-the receiver. Therefore it can be valuable to expose maximum number of allowed,
-currently open and currently used streams to the application to make the mapping
-of data to streams dependent on this information.
+the receiver. Therefore it can be valuable to expose the maximum number of 
+allowed, currently open and currently used streams to the application to make
```
martinthomson, line 331: Oxford comma

draft-ietf-quic-applicability.md
```diff
@@ -421,8 +428,8 @@ QUIC provides an interface that provides multiple streams to the application;
 however, the application usually cannot control how data transmitted over one
 stream is mapped into frames or how those frames are bundled into packets.
 
-By default, many QUIC implementations will try to maximally pack packets with
-one or more stream data frames to minimize bandwidth consumption and
+By default, many QUIC implementations will try to maximally pack QUIC packets
```
martinthomson, line 431: Two lots of QUIC in the same sentence

draft-ietf-quic-applicability.md
britram: commits words and leaves. embarrassing. thanks!

britram:
```suggestion
allowed, currently open, and currently used streams to the application to make
```

draft-ietf-quic-applicability.md
britram:
```suggestion
designing an application's use of streams:
```

draft-ietf-quic-applicability.md
britram:
```suggestion
By default, many implementations will try to maximally pack QUIC packets
```

mirjak (approved):

draft-ietf-quic-applicability.md
```diff
@@ -309,20 +309,27 @@ generate error messages on the application layer to inform the other end and/or
...
+be received in order, that data should be sent on the same stream.
+
+- Multiple streams provide concurrency. Data that can be processed
+independently, and therefore would suffer from head of line blocking if forced
+to be received in order, should be transmitted over separate streams.
+
+- Streams can provide message orientation, and allow messages to be cancelled.
+If one message is mapped to a single stream, resetting the stream to expire an
+unacknowledged message can be used to emulate partial reliability
+on a message basis.
```
gorryfair, line 325: /a message basis/ ???
- could be: /for a specific message/

draft-ietf-quic-applicability.md
```diff
@@ -309,20 +309,27 @@ generate error messages on the application layer to inform the other end and/or
...
+
+If a QUIC receiver has opened the maximum allowed concurrent
+streams, and the sender indicates that more streams are needed, it
 doesn't automatically lead to an increase of the maximum number of streams by
-the receiver. Therefore it can be valuable to expose maximum number of allowed,
-currently open and currently used streams to the application to make the mapping
-of data to streams dependent on this information.
+the receiver. Therefore it can be valuable to expose the maximum number of
+allowed, currently open, and currently used streams to the application to make
+the mapping of data to streams dependent on this information.
```
gorryfair, line 332: Bizarrely-complex sentence? 
Perhaps something like:
"Therefore, an application can utilise the maximum number of allowed, currently open, and currently used streams when determining how to map data to streams."

draft-ietf-quic-applicability.md
```diff
@@ -484,29 +491,29 @@ identification.
...
+after path migration. Therefore migration usually has a performance impact.
 
-Probing packets, which cannot carry application data, can be sent on multiple
-paths at once. Probing packets can be used to perform address validation,
-measure path characteristics as input for the switching decision, or prime the
-congestion controller in preparation for switching to the new path.
+QUIC probing packets, which cannot carry application data, can be sent on
+multiple paths at once. Probing packets can be used to perform address
+validation, measure path characteristics as input for the switching decision,
+or prime the congestion controller in preparation for switching to the new path.
```
gorryfair, line 516: /prime/ seems like a topic that has some semantics that needs CC discussion, and I don't know where I see this in RFCs. This is discussed issue #197.

gorryfair: Although I see PR #122 used different language and avoided having to discuss this:-).

gorryfair (commented):
See specific comments

draft-ietf-quic-applicability.md
mirjak: This change was just changing the line alignment not the words. Let's keep discussion in #197 for this topic.

draft-ietf-quic-applicability.md
mirjak: Your proposal is missing the important point that a QUIC implementation should actually expose this information.

draft-ietf-quic-applicability.md
mirjak, line 325:
```suggestion
for that message.
```

draft-ietf-quic-applicability.md
gorryfair: That was evident in the text. Try: "Therefore, a QUIC implementation ought to expose the maximum number of allowed, currently open, and currently used streams to help an application determine how to map data to streams."

draft-ietf-quic-applicability.md
britram: on a per-message basis?

britram: (Mirja's suggestion accepted)

draft-ietf-quic-applicability.md
britram: We're pointing out that this is valuable; upgrading this to a "should" might be more than editorial. @mirjak will create a separate PR to discuss.

PR #226, opened 2021-02-12
Make it clear that the spin bit generates samples

### LINKED ISSUE #180: Sect 3.7.2 on spinning to understand the RTT
The text describes an example for measuring the average RTT. I like that, but I think we should call out a little that the mean RTT is only one metric. When looking at management info, I'm often interested in min/max and variance...

So: “A simple linear smoothing... to get a more stable estimate”. Quite so for a table average RTT. I’d prefer it if we could say something about RTT variance at this point. To me it is important that RTT measurements are clear, and we should never assume in a Spec that RTT is something specific, but more generally an estimate that needs to consider the variance.

britram:
...that must be postprocessed to become metrics. Fixes #180.


mirjak (approved):

mirjak:
Waiting for more reviews, e.g. from @gorryfair, before we can merge.

PR #228, opened 2021-02-15
Ordered and reliable will be assumed

### LINKED ISSUE #215: Streams section could discuss in-order delivery assumption
Though the specification does not insist upon it, most implementations will assume that stream data is reliably delivered in order.  This means that senders will send until every byte is delivered or they decide to reset the stream (possibly based on STOP_SENDING).  And it means that receivers won't make data available until a contiguous span of bytes starting at the last unread offset are available.

Implementations might act differently, but variations from this behaviour would need to be agreed by both endpoints.

martinthomson:
Beware those who want to do something else.

Closes #215.

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -369,7 +369,7 @@ decision from the reliability level of the stream.
 ## Ordered and Reliable Delivery
 
 QUIC streams enable ordered and reliable delivery.  Though it is possible for an
-implementation to provide some options that use streams for partial reliability
+implementation to provide options that use streams for partial reliability
 or out-of-order delivery, most implementations will assume that data is
 reliably delivered in order.
 
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -366,6 +366,24 @@ might not be desirable. For such streams, QUIC could either provide an
 explicit interface to control prioritization, or derive the prioritization
 decision from the reliability level of the stream.
 
+## Ordered and Reliable Delivery
+
+QUIC streams enable ordered and reliable delivery.  Though it is possible for an
+implementation to provide some options that use streams for partial reliability
+or out-of-order delivery, most implementations will assume that data is
+reliability delivered in order.
+
+Under this assumption, an endpoint that receives stream data might not make
+forward progress until data that is contiguous with the start of a stream is
+available.  In particular, the receiver might withhold flow control credit until
+contiguous data is delivered to the application.  To support this receive logic,
+an endpoint that sends stream data will attempt to send data from the start of
+the stream until that data is acknowledged.
+
+An endpoint that uses a different sending behaviour and does not negotiate that
+change with its peer might encounter performance issues, especially those that
+related to flow control.
+
 ## Flow Control Deadlocks {#flow-control-deadlocks}
 
 Flow control provides a means of managing access to the limited buffers
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -371,18 +371,17 @@ decision from the reliability level of the stream.
 QUIC streams enable ordered and reliable delivery.  Though it is possible for an
 implementation to provide some options that use streams for partial reliability
 or out-of-order delivery, most implementations will assume that data is
-reliability delivered in order.
+reliably delivered in order.
 
 Under this assumption, an endpoint that receives stream data might not make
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-applicability.md
```diff
@@ -366,6 +366,24 @@ might not be desirable. For such streams, QUIC could either provide an
 explicit interface to control prioritization, or derive the prioritization
 decision from the reliability level of the stream.
 
+## Ordered and Reliable Delivery
+
+QUIC streams enable ordered and reliable delivery.  Though it is possible for an
+implementation to provide some options that use streams for partial reliability
+or out-of-order delivery, most implementations will assume that data is
+reliability delivered in order.
```
janaiyengar, line 374:
```suggestion
reliably delivered in order.
```

draft-ietf-quic-applicability.md
```diff
@@ -366,6 +366,24 @@ might not be desirable. For such streams, QUIC could either provide an
...
+
+Under this assumption, an endpoint that receives stream data might not make
+forward progress until data that is contiguous with the start of a stream is
+available.  In particular, the receiver might withhold flow control credit until
```
janaiyengar, line 378:
```suggestion
available.  In particular, a receiver might withhold flow control credit until
```

draft-ietf-quic-applicability.md
```diff
@@ -366,6 +366,24 @@ might not be desirable. For such streams, QUIC could either provide an
...
+contiguous data is delivered to the application.  To support this receive logic,
+an endpoint that sends stream data will attempt to send data from the start of
+the stream until that data is acknowledged.
```
janaiyengar, line 381: This isn't clear -- maybe rephrase? "until that data" is unclear -- this doesn't mean data from the start of the stream, does it?

draft-ietf-quic-applicability.md
```diff
@@ -366,6 +366,24 @@ might not be desirable. For such streams, QUIC could either provide an
...
+
+An endpoint that uses a different sending behaviour and does not negotiate that
+change with its peer might encounter performance issues, especially those that
+related to flow control.
```
janaiyengar, line 385: It's worse -- the sender can easily get deadlocked, especially if it assumes partial reliability when the receiver does not.

janaiyengar (approved):

draft-ietf-quic-applicability.md
```diff
@@ -366,6 +366,23 @@ might not be desirable. For such streams, QUIC could either provide an
 explicit interface to control prioritization, or derive the prioritization
 decision from the reliability level of the stream.
 
+## Ordered and Reliable Delivery
+
+QUIC streams enable ordered and reliable delivery.  Though it is possible for an
+implementation to provide some options that use streams for partial reliability
```
ianswett, line 372:
```suggestion
implementation to provide options that use streams for partial reliability
```

draft-ietf-quic-applicability.md
```diff
@@ -366,6 +366,23 @@ might not be desirable. For such streams, QUIC could either provide an
...
+or out-of-order delivery, most implementations will assume that data is
+reliably delivered in order.
+
+Under this assumption, an endpoint that receives stream data might not make
+forward progress until data that is contiguous with the start of a stream is
+available.  In particular, a receiver might withhold flow control credit until
+contiguous data is delivered to the application.  To support this receive logic,
```
ianswett, line 379: This last sentence seems somewhat duplicative of text in transport, and given it's fairly proscriptive, I'd prefer to remove it,

mirjak:
Isn't this already covered by the next section on "Flow Control Deadlocks"? If not, what's missing? And can we just add a sentence or two there?

draft-ietf-quic-applicability.md
martinthomson: You mean the "in particular"?  Would adding a reference to Section 2.2 of Transport help?  The text in transport only says that implementations MUST be able to deliver in order; it doesn't really say anything about the consequences of that, which is what this is emphasizing.

martinthomson (commented):
@mirjak, the next section talks about deadlocking more generally, on the assumption that data is delivered in order; the deadlock mentioned here is an extra special one.  (I thought that the mention of deadlocks was a nice segue, as it happens.)

draft-ietf-quic-applicability.md
ianswett: A reference would be helpful, but good point on emphasis, happy to keep it in.

PR #233, opened 2021-02-15
Recommendation on ACKing

mirjak:
This is text from Gorry. I created a new section in the applicability document but not fully sure yet where this belongs.

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -449,7 +449,7 @@ padding.
 # ACK-only packets on constrained links
 
 On network paths where there is significant asymmetry,
-the transmission of QUIC ACKs can reduce the efficient use of network
+the transmission of packets only containing ACK frames can reduce the efficient use of network
 capacity or transmission opportunities. This effect occurs
 when the return capacity is significantly more constrained than the
 forward capacity, or the cost of transmission per packet is a
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -450,7 +450,7 @@ padding.
 
 On network paths where there is significant asymmetry,
 the transmission of packets only containing ACK frames can reduce the efficient use of network
-capacity or transmission opportunities. This effect occurs
+capacity. This effect occurs
 when the return capacity is significantly more constrained than the
 forward capacity, or the cost of transmission per packet is a
 significant component of the total transmission cost. Reducing
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -448,15 +448,11 @@ padding.
 
 # ACK-only packets on constrained links
 
-On network paths where there is significant asymmetry,
-the transmission of packets only containing ACK frames can reduce the efficient use of network
-capacity. This effect occurs
-when the return capacity is significantly more constrained than the
-forward capacity, or the cost of transmission per packet is a
-significant component of the total transmission cost. Reducing
-the ratio of acknowledgements to data at the QUIC receiver can improve
-link utilization and reduce link transmission costs.
-
+The cost of sending acknowledgments - in processing cost or link
+utilization - could be a significant proportion of available resources if
+these resources are constrained. Reducing the rate at which acknowledgments
+are generated can preserve these resources and improve overall performance,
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-applicability.md
```diff
@@ -446,6 +446,17 @@ leakage of information about the data that is sent. A QUIC implementation can
 expose an interface that allows an application layer to specify how to apply
 padding.
 
+# ACKing on constraint links
```
ianswett, line 449:
```suggestion
# ACK-only packets on constrained links
```

draft-ietf-quic-applicability.md
```diff
@@ -446,6 +446,17 @@ leakage of information about the data that is sent. A QUIC implementation can
...
+
+On network paths where there is significant asymmetry,
+the transmission of QUIC ACKs can reduce the efficient use of network
```
ianswett, line 452:
```suggestion
the transmission of packets only containing ACK frames can reduce the efficient use of network
```

draft-ietf-quic-applicability.md
```diff
@@ -446,6 +446,17 @@ leakage of information about the data that is sent. A QUIC implementation can
...
+capacity or transmission opportunities. This effect occurs
```
ianswett, line 453:
```suggestion
capacity. This effect occurs
```
This sentence is still a bit awkward, but I'm not sure what else to suggest.

draft-ietf-quic-applicability.md
```diff
@@ -446,6 +446,17 @@ leakage of information about the data that is sent. A QUIC implementation can
...
+when the return capacity is significantly more constrained than the
```
ianswett, line 454: I don't believe return and forward are used elsewhere, but this sentence also seems a bit redundant to the previous sentence which discusses asymmetric paths.

ianswett (commented):
I'm not sure this belongs in applicability.  This seems like a transport feature, and I would not expect most applications to have much control over it?

draft-ietf-quic-applicability.md
martinthomson: Maybe instead:

> The cost of sending acknowledgments - in processing cost or link utilization - could be a significant proportion of available resources if these resources are constrained.  Reducing the rate at which acknowledgments are generated can preserve these resources and improve overall performance.

draft-ietf-quic-applicability.md
gorryfair: ... Radio technologies can have costs per-packet, not per byte - and a sending packet high rate impacts that... in ways for QUIC that would not happen currently with TCP. 
I suggest a new additional sentence, something like:
"On some types of link, the rate of transmission of QUIC packets carrying control information can also impact the overall link efficiency and ultimately the user experience."

draft-ietf-quic-applicability.md
gorryfair: This text (or similar) seems OK to me.

Is "overall" here meaning QUIC user performance  (or overall link-layer?) I think impact on performance for the QUIC user is the more significant additional take-away.

martinthomson: Overall was meant to be all-inclusive, which definitely includes application-relevant metrics like time to first byte, total transfer time, transfer rate, ....  It's clear that there are many ways in which this might help.

gorryfair: We agree. Although my question is, it is obvious to a QUIC user that there user performance can be impacted - for several decades they could have used TCP, where other mitigations exist? So can we add "(including application-relevant metrics)" or something?

mirjak:
Okay. I updated the PR. Can people check again? But I think this is ready to merge (after the meeting).

mirjak:
One more question: usually the ack rate is rather controller by the quic stack than the application. Should we add another sentence saying that an interface for indicating the maximum ack rate would be desirable....?

ianswett (approved):

gorryfair:
I think the text we have might be minimal, but is ok. 

(A good app needs to be aware of this or choose to use a good stack. Since this is a transport feature that isn't always required by the spec, then the app designer needs either the stack to work appropriately and/or to allow some control.)

PR #250, opened 2021-02-16
Fix section reference to QUIC-http ALPN

### LINKED ISSUE #217: ALPN doesn't enable use of other ports
> For example, HTTP/3 ([QUIC-HTTP] Sections 3.2 and 3.3) specifies the use of ALPN [RFC7301] for service discovery which allows the server to use and announce a different port number. 

I think that this refers to Alt-Svc (RFC 7838), which is mentioned in [Section 3.1.1](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#name-http-alternative-services).  You might also mention the not-yet-done SVCB record, but that might not be a good idea if you like the idea of publishing this.

mirjak:
fixes #217


MikeBishop:
It still says ALPN, though.  ALPN doesn't have anything to do with using different ports; it has to do with using different protocols on the same port.

mirjak:
Yes, Martin already pointed this out. Fixed now. Do we need to say still more about ALPN?

MikeBishop (approved):

PR #251, opened 2021-02-16
Update version info

### LINKED ISSUE #231: Dated information about version numbers
> Many QUIC versions that start with 0xff implement IETF drafts. QUIC versions that start with 0x0000 are reserved for IETF consensus documents. 

The first is not true really.  There are 16 million versions that start with 0xff and we only published 34 drafts and implemented far fewer than that.  The second is no longer true as we removed the reservation from the registry.  Maybe just strike these sentences.

mirjak:
fixes #231

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -168,11 +168,11 @@ The following information is exposed in QUIC packet headers:
   identifies the version used for that packet. During Version
   Negotiation (see {{version}} and Section 17.2.1 of {{QUIC-TRANSPORT}}), the
   version number field has a special value (0x00000000) that identifies the
-  packet as a Version Negotiation packet. Many QUIC versions that start with
-  0xff implement IETF drafts. QUIC versions that start with 0x0000 are reserved
-  for IETF consensus documents. For example, QUIC version 1 uses version
-  0x00000001.  Operators should expect to observe packets with other version
-  numbers as a result of various internet experiments and future standards.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -168,11 +168,11 @@ The following information is exposed in QUIC packet headers:
...
   Negotiation (see {{version}} and Section 17.2.1 of {{QUIC-TRANSPORT}}), the
   version number field has a special value (0x00000000) that identifies the
-  packet as a Version Negotiation packet. Many QUIC versions that start with
-  0xff implement IETF drafts. QUIC versions that start with 0x0000 are reserved
-  for IETF consensus documents. For example, QUIC version 1 uses version
-  0x00000001.  Operators should expect to observe packets with other version
-  numbers as a result of various internet experiments and future standards.
+  packet as a Version Negotiation packet. Upon time of publishing of this
+  document, QUIC versions that start with 0xff implement IETF drafts. QUIC
+  version 1 uses version 0x00000001. Operators should expect to observe
```
martinthomson, line 173:
```suggestion
  packet as a Version Negotiation packet. QUIC
  version 1 uses version 0x00000001. Operators should expect to observe
```

martinthomson (commented):
Time of publication statements are not generally useful in this context.

draft-ietf-quic-manageability.md
britram: Will merge this as is for editorial changes. Will create separate PR for content change.

PR #252, opened 2021-02-16
clarify 10 0-RTT packets in first round

### LINKED ISSUE #235: 0-RTT is not limited by the initial congestion window
> The amount of 0-RTT protected data is limited by the initial congestion window, typically around 10 packets [RFC6928].

However there are cases where the congestion window can increase when 0-RTT continues to be sent.  Some involve packet loss and so probably won't result in a net increase in the window, though the total number of packets might still exceed the initial window size.  Others, like TLS HelloRetryRequest or large certificates, involve additional round trips in the handshake.

The total amount of 0-RTT is also limited by remembered flow control values, though this is indirect as this counts stream data only and packets might be further padded.

mirjak:
fixes #235


martinthomson (approved):
qualification works

ianswett (approved):
You can't ACK 0-RTT packets in 0-RTT, so realistically it's probably 10 packets worth of 0-RTT data at most, but one could construct some edge cases so this is likely better.

PR #253, opened 2021-02-16
Integrity check and on-path modifications

### LINKED ISSUE #236: Split sentence on integrity protection
> Therefore, devices on path cannot alter any information or bits in QUIC packet headers, except specific parts of Initial packets, since alteration of header information will lead to a failed integrity check at the receiver, and can even lead to connection termination.

This mixes two things: damage to packets that cause integrity protection to fail, and modification of Initial packets.

How about:

> Therefore, devices on path cannot alter any information or bits in QUIC packets, as a failed integrity check as a result of alterations to a packet will cause the receiver to discard the packet.  Some parts of Initial packets can be altered by removing and re-applying the authenticated encryption.  However, the cryptographic handshake validates many fields and any modifications in those fields will result in connection establishment failing.

mirjak:
fixes #236

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -438,9 +438,9 @@ protected. Further, information that was sent and exposed in handshake packets
 sent before the cryptographic context was established are validated later during
 the cryptographic handshake. Therefore, devices on path cannot alter any
 information or bits in QUIC packets. Such alterations would cause the integrity
-check at the end point to fail and as a result the receiver to discard the packet.
+check to fail, which results in the receiver discarding the packet.
 Some parts of Initial packets could be altered by removing and re-applying the
-authenticated encryption which immediate discard at the receiver. However,
+authenticated encryption without immediate discard at the receiver. However,
 the cryptographic handshake validates most fields and any modifications in
 those fields will result in connection establishment failing later on.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -436,10 +436,13 @@ As soon as the cryptographic context is established, all information in the QUIC
 header, including exposed information, is integrity
 protected. Further, information that was sent and exposed in handshake packets
 sent before the cryptographic context was established are validated later during
-the cryptographic handshake.  Therefore, devices on path cannot alter any
-information or bits in QUIC packet headers, except specific parts of Initial
-packets, since alteration of header information  will lead to a failed integrity
-check at the receiver, and can even lead to connection termination.
+the cryptographic handshake. Therefore, devices on path cannot alter any
+information or bits in QUIC packets. Such alterations would cause the integrity
+check at the end point to fail and as a result the receiver to discard the packet.
```
martinthomson, line 441:
```suggestion
check to fail, which results in the receiver discarding the packet.
```

draft-ietf-quic-manageability.md
```diff
@@ -436,10 +436,13 @@ As soon as the cryptographic context is established, all information in the QUIC
...
+Some parts of Initial packets could be altered by removing and re-applying the
+authenticated encryption which immediate discard at the receiver. However,
```
martinthomson, line 443:
```suggestion
authenticated encryption. However,
```

martinthomson (approved):

draft-ietf-quic-manageability.md
mirjak, line 443:
```suggestion
authenticated encryption without immediate discard at the receiver. However,
```

PR #255, opened 2021-02-16
might eventually be able to associate the version...

### LINKED ISSUE #239: Maybe you can link flows sometimes, but not always
> Negotiated version cannot be identified for flows for which a handshake is not observed, such as in the case of connection migration; however, these flows can be associated with flows for which a version has been identified; see Section 3.4.

This implies that the version of QUIC being used in a given flow can be determined simply by linking that flows to one with a handshake.  That assumes that all the work we did to make flows difficult to link was for naught.  Maybe "can" should be a "might" or something along those lines.

I don't really like the implication that the attempt be made to link flows like this either.

mirjak:
fixes #239

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -552,8 +552,8 @@ packet with a given version from a client to which a server responds with an
 Initial packet with the same version implies acceptance of that version.
 
 Negotiated version cannot be identified for flows for which a handshake is not
-observed, such as in the case of connection migration; however, these flows
-might eventually be associated with flows for which a version has been
+observed, such as in the case of connection migration; however, it might be
+possible to associate a flow with a flow for which a version has been
 identified; see {{sec-flow-association}}.
 
 This document focuses on QUIC Version 1, and this section applies only to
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -552,9 +552,9 @@ packet with a given version from a client to which a server responds with an
 Initial packet with the same version implies acceptance of that version.
 
 Negotiated version cannot be identified for flows for which a handshake is not
-observed, such as in the case of connection migration; however, these flows can
-be associated with flows for which a version has been identified; see
-{{sec-flow-association}}.
+observed, such as in the case of connection migration; however, these flows
+might eventually be associated with flows for which a version has been
```
martinthomson, line 556:
```suggestion
observed, such as in the case of connection migration; however, it might be
possible to associate a flow with a flow for which a version has been
```

Eventually implies an inevitability, which still implies that our linkability protection is ineffectual.

mirjak: If you only switch CID (and maybe one IP address) but keep sending data with the same speed to the same endpoint on the same destination port, I would actually say the linkability protection is quite ineffectual.

ianswett (approved):

PR #270, opened 2021-02-17
Stream ID interactions

### LINKED ISSUE #263: QUIC doesn't necessarily expose Stream IDs?
The ops draft currently says:

> While a QUIC implementation must necessarily provide a way for an application
to send data on separate streams, it does not necessarily expose stream
identifiers to the application (see, for example, {{QUIC-HTTP}}, Section 6)
either at the sender or receiver end, so applications should not assume access
to these identifiers.

I don't see anything in Section 6 of HTTP/3 to support that claim.  In fact, HTTP/3 uses Stream IDs in the GOAWAY frame to indicate the last request the server processed; both client and server will require access to the identifiers of streams, or that information is impossible to generate and useless to receive.

The constraint that I tried to use when writing the HTTP draft was that the application layer has no control over the assignment of Stream IDs.  However, once it has requested and been given a stream, it can retrieve the ID of that stream and use it to reference the stream at the application layer.

By the same token, the application layer shouldn't assume it understands the relationship between stream types and Stream IDs; that may change in a future QUIC version, and it should rely on querying the stream properties to identify types.

MikeBishop:
Fixes #263.

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -324,11 +324,21 @@ the receiver. Therefore it can be valuable to expose maximum number of allowed,
 currently open and currently used streams to the application to make the mapping
 of data to streams dependent on this information.
 
-While a QUIC implementation must necessarily provide a way for an application
-to send data on separate streams, it does not necessarily expose stream
-identifiers to the application (see, for example, {{QUIC-HTTP}}, Section 6)
-either at the sender or receiver end, so applications should not assume access
-to these identifiers.
+QUIC assigns a numerical identifier to each stream, called the Stream ID.  While
+the relationship between these identifiers and stream types is clearly defined
+in version 1 of QUIC, future versions might change this relationship for various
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
That captures it.

ianswett (approved):

mirjak (approved):

mikkelfj:
I believe it is sane to not assume a specific ID ordering, but I do not think it is right to require that all applications do not so. You can easily have applications that benefit from knowing and controlling the ID sequencing in detail, but such applications may not be implementable on the majority of QUIC library implementations. Ops should only make this a guidance for designing new app level protocols, not a requirement.

mirjak:
I would see everything in the ops docs as a recommendation and not a requirement. However this PR is already merged. Can you open a new issue (and ideally propose some wording or say more specifically where you think the current wording is too restrictive)?

mikkelfj:
@mirjak done

Issue #242, opened 2021-02-16
Forcing address stability


martinthomson:
This section seems to be saying two things about a NAT that attempts to rewrite addresses to keep them stable:

1. It won't work, because the NAT needs to know about all connection IDs, including those that have been negotiated without its knowledge.
2. Even if it did work, which would require cooperation with endpoints, it would facilitate amplification attacks.

This is good, but I found it unclear.  It took a few readings to figure it out.

gorryfair:
To what text does #242 relate?

martinthomson:
My apologies.  I managed to edit that out.  [Section 4.8.2](https://quicwg.org/ops-drafts/draft-ietf-quic-manageability.html#name-helping-with-routing-infras).

mirjak:
I made a larger rewrite in PR #257. Waiting for a review from @martinduke, however, I actually have couple of questions and I think there is still some redundancy....

Issue #241, opened 2021-02-16
Are endpoints the ones disabling security?


martinthomson:
> In fact, NAT behavior that relies on CID may instead cause connection failure when endpoints change Connection ID, and disable important protocol security features. 

I don't think so, but the structure of [this sentence](https://quicwg.org/ops-drafts/draft-ietf-quic-manageability.html#section-4.8-1) makes it hard to tell.

mirjak:
We left this open as we didn't merge the bigger NAT section rewrite yet.

PR #272, opened 2021-02-18
Rewrite NAT section to minimum text

mirjak:
Note that the main point about using CID in a NAT can break connectivity was already covered the section "Stateful Treatment of QUIC Traffic". The new section now only covers the NAT for routing stability case.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -827,23 +827,25 @@ connectivity breakage later, in case the connection ID changes.
 Use of address rewriting to ensure routing stablity as an approach to
 simplify operational routing conceals client address changes and will
 therefore mask important signals that drive security mechanisms, and
-as auch opens QUIC up to various attacks.
+as such opens QUIC up to various attacks.
 
 While QUIC's migration capability makes it possible for an server to survive
 address changes, this does not work if the routers or switches in the server
-infrastructure rely on address-port 4-tuple as a NAT rebinding or address
+infrastructure route using the address-port 4-tuple. If infrastructure routes on
+addresses only, NAT rebinding or address
 migration will cause packets to be delivered to the wrong server. {{QUIC_LB}}
 described a way to addresses this problem by coordinating the selection and
 use of connection IDs between load-balancers and servers.
 
-An alternative, potentially simpler approach appears to be the use of NAT
-in front of such an infrastructure setup. However, hiding information about the
-change of the IP address or port conceals important and security relevant
+Applying address translation at a middlebox to maintain a stable
+address-port mapping for flows based on connection ID might seem
+like a solution to this problem. However, hiding information about the
+change of the IP address or port conceals important and security-relevant
 information from QUIC endpoints and as such would facilitate amplification
-attacks (see section 9 of {{QUIC-TRANSPORT}}). An NAT function that bleaches
-peer address changes, hinders the other end
-to detect and mitigate these attacks and verify connectivity to the
-new address based on QUIC PATH_CHALLENGE and PATH_RESPONSE frames.
+attacks (see section 9 of {{QUIC-TRANSPORT}}). An NAT function that hides
+peer address changes prevents the other end from
+detecting and mitigating attacks as the endpoint cannot verify connectivity to the
+new address using QUIC PATH_CHALLENGE and PATH_RESPONSE frames.
 
 In addition, a change of IP address or port is also an input signal to other
 internal mechanisms in QUIC. When a path change is detected, path-dependent
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -824,27 +824,26 @@ breakage later on connection ID change.
 
 ## Address Rewriting to Ensure Routing Stability
 
-If the routers or switches in the server infrastructure rely on address-port
-4-tuple, a NAT rebinding or any other address migration on the client side,
-even if that could be handled by the QUIC server, will cause packets to be
-delivered to the wrong server. In this case it seems tempting to use NAT in
-front of the server infrastructure to provide a stable address 4-tuple,
-rather than changing the deployed routing infrastructure. As the connection
-ID can change and not-yet-used values are not observable on path, this would
-require servers cooperate to inform the NAT of all new connection IDs,
-similarly as described in {{QUIC_LB}} for load balancers. However, this
-approach is not recommended to be used with NAT functions, as hiding address
-changes from a server prevents the server from validating new addresses,
-which would facilitate amplification attacks (see
-{{Section 9 of QUIC-TRANSPORT}}).
+Use of address rewriting to ensure routing stablity as an approach to
+simplify operational routing conceals client address changes and will
+therefore mask important signals that drive security mechanisms, and
+as auch opens QUIC up to various attacks.
 
-For example, an attacker might copy a legitimate QUIC packet and change the
-source address to match its own. In the absence of a bleaching NAT, the
-receiving endpoint would interpret this as a potential NAT rebinding and use a
-PATH_CHALLENGE frame to prove that the peer endpoint is not truly at the new
-address, thus thwarting the attack. A bleaching NAT has no means of sending an
-encrypted PATH_CHALLENGE frame, so it might start redirecting all QUIC traffic
-to the attacker address and thus allow an observer to break the connection.
+While QUIC's migration capability makes it possible for an server to survive
+address changes, this does not work if the routers or switches in the server
+infrastructure rely on address-port 4-tuple as a NAT rebinding or address
+migration will cause packets to be delivered to the wrong server. {{QUIC_LB}}
+described a way to addresses this problem by coordinating the selection and
+use of connection IDs between load-balancers and servers.
+
+An alternative, potentially simpler approach appears to be the use of NAT
+in front of such an infrastructure setup. However, hiding information about the
+change of the IP address or port conceals important and security relevant
+information from QUIC endpoints and as such would facilitate amplification
+attacks (see section 9 of {{QUIC-TRANSPORT}}). An NAT function that bleaches
+peer address changes, hinders the other end
+to detect and mitigate these attacks and verify connectivity to the
+new address based on QUIC PATH_CHALLENGE and PATH_RESPONSE frames.
 
 In addition, a change of IP address or port is also an input signal to other
 internal mechanisms in QUIC. When a path change is detected, path-dependent
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -820,7 +820,7 @@ If a NAT hits an operational limit, it is recommended to rather drop the
 initial packets of a flow (see also {{sec-filtering}}),
 which potentially triggers a fallback to TCP, than using the connection ID to
 muliplex mulitple connections on the same IP address port pair which risks
-breakage later on connection ID change.
+connectivity breakage later, in case the connection ID changes.
 
 ## Address Rewriting to Ensure Routing Stability
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -780,6 +780,13 @@ in the downstream direction, and vice versa.
 In this section, we review specific network management and measurement
 techniques and how QUIC's design impacts them.
 
+## Passive Network Performance Measurement and Troubleshooting
+
+Limited RTT measurement is possible by passive observation of QUIC traffic;
+see {{sec-rtt}}. No passive measurement of loss is possible with the present
+wire image. Extremely limited observation of upstream congestion may be
+possible via the observation of CE markings on ECN-enabled QUIC traffic.
+
 ## Stateful Treatment of QUIC Traffic {#sec-stateful}
 
 Stateful treatment of QUIC traffic (e.g., at a firewall or NAT middlebox) is
@@ -809,12 +816,40 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
 of state in the middle of a connection. Use of connection ID specifically
 discouraged for NAT applications.
 
-## Passive Network Performance Measurement and Troubleshooting
+If a NAT hits an operational limit, it is recommended to rather drop the
+initial packets of a flow (see allso {{sec-filtering}}),
+which potentially triggers a fallback to TCP, than using the connection ID to
+muliplex mulitple connections on the same IP address port pair which risks
+breakage later on connection ID change.
 
-Limited RTT measurement is possible by passive observation of QUIC traffic;
-see {{sec-rtt}}. No passive measurement of loss is possible with the present
-wire image. Extremely limited observation of upstream congestion may be
-possible via the observation of CE markings on ECN-enabled QUIC traffic.
+## Address Rewriting to Ensure Routing Stability
+
+If the routers or switches in the server infrastructure rely on address-port
+4-tuple, a NAT rebinding or any other address migration on the client side,
+even if that could be handled by the QUIC server, will cause packets to be
+delivered to the wrong server. In this case it seems temping to use NAT in
+front of the server infrastructure to provide a stable address 4-tuple,
+rather than changing the deployed routing infrastructure. As the connection
+ID can change and not-yet-used values are not observable on path, this would
+require servers cooperate to inform the NAT of all new connection IDs,
+similarly as described in {{QUIC_LB}} for load balancers. However, this
+approach is not recommended to be used with NAT functions, as hiding address
+changes from a server prevents the server from validating new addresses,
+which would facilitate amplification attacks (see
+{{Section 9 of QUIC-TRANSPORT}}).
+
+For example, an attacker might copy a legitimate QUIC packet and change the
+source address to match its own. In the absence of a bleaching NAT, the
+receiving endpoint would interpret this as a potential NAT rebinding and use a
+PATH_CHALLENGE frame to prove that the peer endpoint is not truly at the new
+address, thus thwarting the attack. A bleaching NAT has no means of sending an
+encrypted PATH_CHALLENGE frame, so it might start redirecting all QUIC traffic
+to the attacker address and thus allow an observer to break the connection.
+
+In addition, a change of IP address or port is also an input signal to other
+internal mechanisms in QUIC. When a path change is detected, path-dependent
+variables like congestion control parameters will be reset protecting
+the new path from overload.
 
 ## Server Cooperation with Load Balancers {#sec-loadbalancing}
 
@@ -822,7 +857,42 @@ In the case of content distribution networking architectures including load
 balancers, the connection ID provides a way for the server to signal information
 about the desired treatment of a flow to the load balancers. Guidance on
 assigning connection IDs is given in
-{{?QUIC-APPLICABILITY=I-D.ietf-quic-applicability}}.
+{{?QUIC-APPLICABILITY=I-D.ietf-quic-applicability}} and {{QUIC_LB}}
+describes a system for coordinating selection and use of connection IDs between
+load-balancers and servers.
+
+## Filtering behavior {#sec-filtering}
+
+{{?RFC4787}} describes possible packet filtering behaviors that relate to NATs.
+Though the guidance there holds, a particularly unwise behavior is to admit a
+handful of UDP packets and then make a decision as to whether or not to filter
+it. QUIC applications are encouraged to fail over to TCP if early packets do
+not arrive at their destination {{?I-D.ietf-quic-applicability}}, as QUIC is
+based on UDP and there are known block of UDP (see Section {{sec-udp-1312}}).
+Admitting a few packets allows the QUIC endpoint to determine that the path
+accepts QUIC. Sudden drops afterwards will result in slow and costly timeouts
+before abandoning the connection.
+
+## UDP Policing {#sec-udp-1312}
+
+Today, UDP is the most prevalent DDoS vector, since it is easy for compromised
+non-admin applications to send a flood of large UDP packets (while with TCP the
+attacker gets throttled by the congestion controller) or to craft reflection and
+amplification attacks. Networks should therefore be
+prepared for UDP flood attacks on ports used for QUIC traffic. One possible
+response to this threat is to police UDP traffic on the network, allocating a
+fixed portion of the network capacity to UDP and blocking UDP datagram over that
+cap.
+
+The recommended way to police QUIC
+packets is to either drop them all or to throttle them based on the hash of the
+UDP datagram's source and destination addresses, blocking a portion of the hash
+space that corresponds to the fraction of UDP traffic one wishes to drop.
+When the handshake is blocked, QUIC-capable applications may failover to TCP
+(at least applications using well-known UDP ports). However, blindly blocking a
+significant fraction of QUIC packets will allow many QUIC handshakes
+to complete, preventing a TCP failover, but the connections will suffer from
+severe packet loss.
 
 ## DDoS Detection and Mitigation {#sec-ddos-dec}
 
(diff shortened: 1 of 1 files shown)


mirjak:
Note also that this PR also contains some rearrangement of sections.

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
 of state in the middle of a connection. Use of connection ID specifically
 discouraged for NAT applications.
 
-## Passive Network Performance Measurement and Troubleshooting
+If a NAT hits an operational limit, it is recommended to rather drop the
+initial packets of a flow (see allso {{sec-filtering}}),
```
martinthomson, line 820:
```suggestion
initial packets of a flow (see also {{sec-filtering}}),
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+which potentially triggers a fallback to TCP, than using the connection ID to
+muliplex mulitple connections on the same IP address port pair which risks
+breakage later on connection ID change.
+
+## Address Rewriting to Ensure Routing Stability
+
+If the routers or switches in the server infrastructure rely on address-port
+4-tuple, a NAT rebinding or any other address migration on the client side,
+even if that could be handled by the QUIC server, will cause packets to be
+delivered to the wrong server. In this case it seems temping to use NAT in
```
martinthomson, line 830: This sentence is too long.

martinthomson, line 830:
```suggestion
If the routers or switches in the server infrastructure rely on address-port
4-tuple, a NAT rebinding or any other address migration on the client side,
even if that could be handled by the QUIC server, will cause packets to be
delivered to the wrong server. In this case it seems tempting to use NAT in
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
-wire image. Extremely limited observation of upstream congestion may be
-possible via the observation of CE markings on ECN-enabled QUIC traffic.
+For example, an attacker might copy a legitimate QUIC packet and change the
+source address to match its own. In the absence of a bleaching NAT, the
+receiving endpoint would interpret this as a potential NAT rebinding and use a
+PATH_CHALLENGE frame to prove that the peer endpoint is not truly at the new
+address, thus thwarting the attack. A bleaching NAT has no means of sending an
+encrypted PATH_CHALLENGE frame, so it might start redirecting all QUIC traffic
+to the attacker address and thus allow an observer to break the connection.
+
```
martinthomson, line 848:
```suggestion
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+the new path from overload.
 
 ## Server Cooperation with Load Balancers {#sec-loadbalancing}
 
 In the case of content distribution networking architectures including load
 balancers, the connection ID provides a way for the server to signal information
 about the desired treatment of a flow to the load balancers. Guidance on
 assigning connection IDs is given in
-{{?QUIC-APPLICABILITY=I-D.ietf-quic-applicability}}.
+{{?QUIC-APPLICABILITY=I-D.ietf-quic-applicability}} and {{QUIC_LB}}
```
martinthomson, line 860:
```suggestion
{{?QUIC-APPLICABILITY=I-D.ietf-quic-applicability}}. {{QUIC_LB}}
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+describes a system for coordinating selection and use of connection IDs between
+load-balancers and servers.
+
+## Filtering behavior {#sec-filtering}
```
martinthomson, line 864:
```suggestion
## Filtering Behavior {#sec-filtering}
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+
+{{?RFC4787}} describes possible packet filtering behaviors that relate to NATs.
+Though the guidance there holds, a particularly unwise behavior is to admit a
+handful of UDP packets and then make a decision as to whether or not to filter
+it. QUIC applications are encouraged to fail over to TCP if early packets do
+not arrive at their destination {{?I-D.ietf-quic-applicability}}, as QUIC is
+based on UDP and there are known block of UDP (see Section {{sec-udp-1312}}).
```
martinthomson, line 871:
```suggestion
based on UDP and there are known block of UDP (see {{sec-udp-1312}}).
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+non-admin applications to send a flood of large UDP packets (while with TCP the
+attacker gets throttled by the congestion controller) or to craft reflection and
+amplification attacks. Networks should therefore be
+prepared for UDP flood attacks on ports used for QUIC traffic. One possible
+response to this threat is to police UDP traffic on the network, allocating a
+fixed portion of the network capacity to UDP and blocking UDP datagram over that
+cap.
+
+The recommended way to police QUIC
+packets is to either drop them all or to throttle them based on the hash of the
```
martinthomson, line 888:
```suggestion
packets is to either drop them all or to throttle them based on
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+UDP datagram's source and destination addresses, blocking a portion of the hash
```
martinthomson, line 889:
```suggestion
only source and destination addresses.
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+space that corresponds to the fraction of UDP traffic one wishes to drop.
```
martinthomson, line 890:
```suggestion
```

The extra words don't really help here.  If anything, you need to highlight the need to block only a subset of addresses rather than a subset of packets.

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+When the handshake is blocked, QUIC-capable applications may failover to TCP
+(at least applications using well-known UDP ports). However, blindly blocking a
+significant fraction of QUIC packets will allow many QUIC handshakes
```
martinthomson, line 893:
```suggestion
(at least applications using well-known UDP ports). However, blocking a
fraction of QUIC packets will allow many QUIC handshakes
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+to complete, preventing a TCP failover, but the connections will suffer from
```
martinthomson, line 894:
```suggestion
to complete, which might discouraging endpoints from attempting to fall back to TCP.
The increased loss the connections experience will only degrade performance.
```

martinthomson (commented):
This is better, I think.

draft-ietf-quic-manageability.md
mirjak: Okay for me to take this example text out.

draft-ietf-quic-manageability.md
mirjak: I know. That's my German gens come through. I will break it up in two sentences.

draft-ietf-quic-manageability.md
mirjak: Yes, that was the intention here and I think your proposed edit changes the meaning.

(Also I only moved this part of the text in this PR but didn't change the wording. Maybe a separate PR would be better here.)

draft-ietf-quic-manageability.md
mirjak: This also changes the meaning completely. This sentence was supposed to say that if you instead just blindly block random QUIC packets without being flow-aware this is really bad.

draft-ietf-quic-manageability.md
mirjak: I think prevent fallback is more correct here.

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
 of state in the middle of a connection. Use of connection ID specifically
 discouraged for NAT applications.
 
-## Passive Network Performance Measurement and Troubleshooting
+If a NAT hits an operational limit, it is recommended to rather drop the
+initial packets of a flow (see also {{sec-filtering}}),
+which potentially triggers a fallback to TCP, than using the connection ID to
+muliplex mulitple connections on the same IP address port pair which risks
+breakage later on connection ID change.
```
mirjak, line 823:
```suggestion
connectivity breakage later, in case the connection ID changes.
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+to the attacker address and thus allow an observer to break the connection.
+
+In addition, a change of IP address or port is also an input signal to other
+internal mechanisms in QUIC. When a path change is detected, path-dependent
+variables like congestion control parameters will be reset protecting
+the new path from overload.
 
 ## Server Cooperation with Load Balancers {#sec-loadbalancing}
 
 In the case of content distribution networking architectures including load
```
ianswett, line 856: There are a lot of server architectures which are not CDNs and may want to use something like QUIC-LB.

```suggestion
In the case of networking architectures including load
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
 of state in the middle of a connection. Use of connection ID specifically
 discouraged for NAT applications.
 
-## Passive Network Performance Measurement and Troubleshooting
+If a NAT hits an operational limit, it is recommended to rather drop the
```
ianswett, line 819: I think the caution against using Connection ID needs to be a lot stronger.  Things will break, users will notice.  I realize that implementations have to use this to enforce that, but I believe some already are.  See a comment below as well.

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+Though the guidance there holds, a particularly unwise behavior is to admit a
+handful of UDP packets and then make a decision as to whether or not to filter
+it. QUIC applications are encouraged to fail over to TCP if early packets do
+not arrive at their destination {{?I-D.ietf-quic-applicability}}, as QUIC is
+based on UDP and there are known block of UDP (see {{sec-udp-1312}}).
+Admitting a few packets allows the QUIC endpoint to determine that the path
+accepts QUIC. Sudden drops afterwards will result in slow and costly timeouts
+before abandoning the connection.
+
+## UDP Policing {#sec-udp-1312}
```
ianswett, line 876: Policing has a slightly different meaning(token bucket policer) that's widely used, so can we find a different term in this context?

draft-ietf-quic-manageability.md
ianswett:
```suggestion
(at least applications using well-known UDP ports). However, blocking a
random fraction of QUIC packets across 4-tuples will allow many QUIC handshakes
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+Today, UDP is the most prevalent DDoS vector, since it is easy for compromised
+non-admin applications to send a flood of large UDP packets (while with TCP the
+attacker gets throttled by the congestion controller) or to craft reflection and
+amplification attacks. Networks should therefore be
+prepared for UDP flood attacks on ports used for QUIC traffic. One possible
+response to this threat is to police UDP traffic on the network, allocating a
+fixed portion of the network capacity to UDP and blocking UDP datagram over that
+cap.
+
+The recommended way to police QUIC
```
ianswett, line 887: This is recommending something that relies on the 4-tuple, but particularly if there's another NAT involved, the 4-tuple may not be stable.  So this puts connections in a situation where the handshake completes, they use the connection for a bit, a NAT rebind occurs, and then the connection is blackholed.

There are defenses against this problem, but this seems like a fairly dangerous recommendation.

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+accepts QUIC. Sudden drops afterwards will result in slow and costly timeouts
+before abandoning the connection.
+
+## UDP Policing {#sec-udp-1312}
+
+Today, UDP is the most prevalent DDoS vector, since it is easy for compromised
+non-admin applications to send a flood of large UDP packets (while with TCP the
+attacker gets throttled by the congestion controller) or to craft reflection and
+amplification attacks. Networks should therefore be
+prepared for UDP flood attacks on ports used for QUIC traffic. One possible
```
ianswett, line 882: If the port really is only used for QUIC, then we're in fairly good shape, it's if the port is used by QUIC, but isn't QUIC traffic.

ianswett (commented):
There's a lot here, and the tradeoffs are subtle and fairly scary in my opinion.  Some of this may be worth discussion at the meeting.

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+
+If the routers or switches in the server infrastructure rely on address-port
+4-tuple, a NAT rebinding or any other address migration on the client side,
+even if that could be handled by the QUIC server, will cause packets to be
+delivered to the wrong server. In this case it seems tempting to use NAT in
+front of the server infrastructure to provide a stable address 4-tuple,
+rather than changing the deployed routing infrastructure. As the connection
+ID can change and not-yet-used values are not observable on path, this would
+require servers cooperate to inform the NAT of all new connection IDs,
+similarly as described in {{QUIC_LB}} for load balancers. However, this
```
martinduke, line 835: This is not how QUIC-LB works; the servers do not inform the LB about new connection IDs. In fact, there is no way in QUIC-LB to match two connection IDs to the same connection, merely to the same server.

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
-## Passive Network Performance Measurement and Troubleshooting
+If a NAT hits an operational limit, it is recommended to rather drop the
+initial packets of a flow (see also {{sec-filtering}}),
+which potentially triggers a fallback to TCP, than using the connection ID to
+muliplex mulitple connections on the same IP address port pair which risks
+breakage later on connection ID change.
+
+## Address Rewriting to Ensure Routing Stability
+
+If the routers or switches in the server infrastructure rely on address-port
```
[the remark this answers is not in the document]
martinduke: I don't think this section captures the purpose of the text it's replacing.

1.  Whether or not I have QUIC-LB, if there's some part of the infrastructure that wants stable 4tuples,  I could try to use the CID to catch NAT rebindings and just rewrite the source address to mask the change from the infrastructure. Yes, this will break if CIDs change; but even if the CID stays the same it will also break the security mechanisms that protect attacks related to NAT rebinding. This is the point of the original text.

2. Whether or not I have QUIC-LB, if the CID and 4tuple change there is no way to maintain the original source and IP port for the connection.

draft-ietf-quic-manageability.md
mirjak: Blocking?

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,83 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+4-tuple, a NAT rebinding or any other address migration on the client side,
+even if that could be handled by the QUIC server, will cause packets to be
+delivered to the wrong server. In this case it seems tempting to use NAT in
+front of the server infrastructure to provide a stable address 4-tuple,
+rather than changing the deployed routing infrastructure. As the connection
+ID can change and not-yet-used values are not observable on path, this would
+require servers cooperate to inform the NAT of all new connection IDs,
```
mirjak, line 834:
```suggestion
require servers cooperate with the in-network device about the use of connection IDs,
```

Is this okay @martinduke ?

draft-ietf-quic-manageability.md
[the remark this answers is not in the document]
mirjak: Regarding your point 1: I think this is covered by 
"this approach is not recommended to be used with NAT functions, as hiding address
 changes from a server prevents the server from validating new addresses,
 which would facilitate amplification attacks"
What's missing?

Regarding you point 2, I would think it is also possible to use similar mechanisms as described in the LB doc to not only encode with server but also the actual connection, no?

draft-ietf-quic-manageability.md
mirjak: Or should we keep the original text and remove the reference to the LB doc?

draft-ietf-quic-manageability.md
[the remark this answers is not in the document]
mirjak: This is the text from the other rewrite attempt:

```
Concealing client address changes in order to simplify operational routing
 issues will mask important signals that drive security mechanisms, and
 therefore opens QUIC up to various attacks.

While QUIC's migration capability makes it possible for an server to survive
 address changes, this does not work if the routers or switches in the server
 infrastructure rely on address-port 4-tuple as a NAT rebinding or address
 migration will cause packets to be delivered to the wrong server. {{QUIC_LB}}
 addresses this problem proposing a QUIC extension to allow server-load balancer
 coordination to routable CIDs.

An alternative, potentially simpler approach would seem to be the use of NAT
 in front of such an infrastructure setup. However, hiding information about the
 change of the IP address or port conceals important, and security relevant
 information from QUIC endpoints, and as such would facilitate amplification
 attacks (see section 9 of {{QUIC-TRANSPORT}}).

 An NAT that bleaches peer address changes, hinders the other end
 to detect and mitigate these attacks and verify connectivity to the
 new address based on QUIC PATH_CHALLENGE and PATH_RESPONSE frames.
```

Or is this a better starting point?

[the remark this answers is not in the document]
martinduke: I like the other rewrite text better, yes.

On point 1: what I'm saying is that QUIC-LB doesn't enter into it.

On point 2: yes, that is theoretically possible. However, I don't want to suggest in any way that we make CIDs absolutely linkable. Exposing the server mapping is bad enough!

draft-ietf-quic-manageability.md
mirjak: Yes that's what the text is trying to say.

draft-ietf-quic-manageability.md
mirjak: I only copied this section to a different position in the draft. Can you please add a separate issue for this.

draft-ietf-quic-manageability.md
mirjak: Can you open a separate issue for this and propose some text (actually in the issue for now maybe, until this PR is merged) as this PR is only moving but not changing this part of the text.

draft-ietf-quic-manageability.md
mirjak: or just filtering?

mirjak: Maybe this and the previous section should be merged. But let me merge in the PR first and then do another PR if needed.

draft-ietf-quic-manageability.md
mirjak: Also I also musst copied this text to a different place in the draft. Can you maybe open a separate issue for this.

mirjak:
I updated the text in the "Address Rewriting to Ensure Routing Stability" as agreed with @martinduke. The remaining comments are on sections where I didn't change any of the text but did move them to a different position in the document. Please open separate issue for them!

I would really like to merge this PR now/first so we have all the pieces in the right place. We can then create new PRs on further changes needed. Is that okay?

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,82 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+which potentially triggers a fallback to TCP, than using the connection ID to
+muliplex mulitple connections on the same IP address port pair which risks
+connectivity breakage later, in case the connection ID changes.
+
+## Address Rewriting to Ensure Routing Stability
+
+Use of address rewriting to ensure routing stablity as an approach to
+simplify operational routing conceals client address changes and will
+therefore mask important signals that drive security mechanisms, and
+as auch opens QUIC up to various attacks.
```
martinthomson, line 830:
```suggestion
as such opens QUIC up to various attacks.
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,82 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+
+While QUIC's migration capability makes it possible for an server to survive
+address changes, this does not work if the routers or switches in the server
+infrastructure rely on address-port 4-tuple as a NAT rebinding or address
+migration will cause packets to be delivered to the wrong server. {{QUIC_LB}}
```
martinthomson, line 835:
```suggestion
infrastructure route using the address-port 4-tuple. If infrastructure routes on
addresses only, NAT rebinding or address
migration will cause packets to be delivered to the wrong server. {{QUIC_LB}}
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,82 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+described a way to addresses this problem by coordinating the selection and
+use of connection IDs between load-balancers and servers.
+
+An alternative, potentially simpler approach appears to be the use of NAT
+in front of such an infrastructure setup. However, hiding information about the
```
martinthomson, line 840:
```suggestion
Applying address translation at a middlebox to maintain a stable address-port mapping for flows based on connection ID might seem like a solution to this problem. However, hiding information about the
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,82 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+change of the IP address or port conceals important and security relevant
```
martinthomson, line 841:
```suggestion
change of the IP address or port conceals important and security-relevant
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,82 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+information from QUIC endpoints and as such would facilitate amplification
+attacks (see section 9 of {{QUIC-TRANSPORT}}). An NAT function that bleaches
```
martinthomson, line 843:
```suggestion
attacks (see section 9 of {{QUIC-TRANSPORT}}). An NAT function that hides
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,82 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+peer address changes, hinders the other end
```
martinthomson, line 844:
```suggestion
peer address changes prevents the other end from
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,82 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+to detect and mitigate these attacks and verify connectivity to the
```
martinthomson, line 845:
```suggestion
detecting and mitigating attacks as the endpoint cannot verify connectivity to the
```

draft-ietf-quic-manageability.md
```diff
@@ -809,20 +816,82 @@ off the connection ID may therefore cause undetectable and unrecoverable loss
...
+new address based on QUIC PATH_CHALLENGE and PATH_RESPONSE frames.
```
martinthomson, line 846:
```suggestion
new address using QUIC PATH_CHALLENGE and PATH_RESPONSE frames.
```

martinthomson (commented):
Getting better.  Though this PR is now at the point that I can't follow all the different pieces.

mirjak:
addresses #241 and addresses #242

PR #285, opened 2021-03-04
CID must be present and is directional

### LINKED ISSUE #280: linking CIDs?
This is branched of #239.

@martinthomson said:

"However, it needs to specify that the connection ID association needs to be directional. Maybe that is assumed, but it should be stated. And it needs to say that this is unlikely to apply when the connection ID is empty."

mirjak:
fixes #280

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -674,9 +674,10 @@ The QUIC connection ID (see {{rebinding}}) is designed to allow an on-path
 device such as a load-balancer to associate two flows as identified by
 five-tuple when the address and port of one of the endpoints changes; e.g. due
 to NAT rebinding or server IP address migration. An observer keeping flow state
-can associate a connection ID with a given flow, and can associate a known flow
-with a new flow when when observing a packet sharing a connection ID and one
-endpoint address (IP address and port) with the known flow.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -674,9 +674,10 @@ The QUIC connection ID (see {{rebinding}}) is designed to allow an on-path
 device such as a load-balancer to associate two flows as identified by
 five-tuple when the address and port of one of the endpoints changes; e.g. due
 to NAT rebinding or server IP address migration. An observer keeping flow state
-can associate a connection ID with a given flow, and can associate a known flow
-with a new flow when when observing a packet sharing a connection ID and one
-endpoint address (IP address and port) with the known flow.
+can associate a connection ID, if present, with a given flow, and can associate
+a known flow with a new flow when when observing a packet sharing the same
+connection ID in the same direction between client and server and one endpoint
+address (IP address and port) with the known flow.
```
martinthomson, line 680:
```suggestion
connection ID in the same direction between client and server and sharing one
endpoint address (IP address and port) with the known flow.
```

This is a bit hard to follow without the extra word.

martinthomson (approved):

PR #288, opened 2021-03-09
Connections aren't reset, they are closed.

mjoras:
Reset is a directional stream semantic, not a connection one.

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -309,7 +309,7 @@ the ingress direction; these actions are fully independent of each other.
 QUIC does not provide an interface for exceptional handling of any stream.
 If a stream that is critical for an application is closed, the application can
 generate error messages on the application layer to inform the other end and/or
-the higher layer, which can eventually reset the QUIC connection.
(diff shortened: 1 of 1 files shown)


mjoras:
Oh, I see. Yeah maybe abort is better. For me reading this "reset" made me think of TCP connection and or QUIC stream semantics, which is probably not the intention.

draft-ietf-quic-applicability.md
```diff
@@ -309,7 +309,7 @@ the ingress direction; these actions are fully independent of each other.
 QUIC does not provide an interface for exceptional handling of any stream.
 If a stream that is critical for an application is closed, the application can
 generate error messages on the application layer to inform the other end and/or
-the higher layer, which can eventually reset the QUIC connection.
+the higher layer, which can eventually result in closure of the QUIC connection.
```
mirjak, line 312:
```suggestion
the higher layer, which can eventually result in aborting of the QUIC connection.
```

britram, line 312:
```suggestion
the higher layer, which can eventually terminate the QUIC connection.
```

Issue #175, opened 2021-02-01
Discuss application-layer graceful/scheduled close


LPardue:
https://quicwg.org/ops-drafts/draft-ietf-quic-applicability.html#section-8 deals with connection closure and is pretty exhaustive. However, I think it would be improved by discussing how an application protocol might manage graceful connection close in the absence of a transport-layer mechanism. HTTP/3 defines the GOAWAY frame for this purpose, defining the closure signalling mechnanism and the  behaviour of endpoints when in the draining period - especially wrt creating streams.

mirjak:
@LPardue would you be able to propose some text?

LPardue:
I'll try 😀

mirjak:
#298 is merged

PR #298, opened 2021-03-24
Split Connection Closure discussion into two parts

### LINKED ISSUE #297: Better discuss the nuance between transport error codes and application error codes
Mention of error codes is restricted to the section `Connection Closure`, which ends up also talking about STOP_SENDING and RESET_STREAM. I think it would be clearer to have a standalone section that describes the different error code spaces and the xref that in the other sections that deal with stream usage and connection closure.

LPardue:
Fixes #297 and #175.

This ended up much larger than I anticiated but I found the old Connection Closure text quite hard to modify in a way that let me frame the graceful close additions. 

I think this is a net positive structural move, even if we might need to bash the section ordering and contained text.

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -496,9 +496,9 @@ Applications that use QUIC can define their own error detection and signaling
 (see, for example, {{Section 8 of QUIC-HTTP}}). Application errors can affect an
 entire connection or a single stream.
 
-QUIC defines an error code space that is used for error handling at the transport layer. QUIC
-encourages endpoints to use the most specific code, although any applicable code
-is permitted, including generic ones.
+QUIC defines an error code space that is used for error handling at the
+transport layer. QUIC encourages endpoints to use the most specific code,
+although any applicable code is permitted, including generic ones.
 
 Applications using QUIC define an error
 code space that is independent from QUIC or other applications (see, for
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -457,14 +457,14 @@ being terminated in some protocols.
 # Stream Limit Commitments
 
 QUIC endpoints are responsible for communicating the cumulative limit of streams
-they would allow to be opened by their peer. Initial limits are advertised
-using the initial_max_streams_bidi and initial_max_streams_uni transport
-parameters. As streams are opened and closed they are consumed and the cumulative
-total is incremented. Limits can be increased using the MAX_STREAMS frame but
-there is no mechanism to reduce limits. Once stream limits are reached, no more
-streams can be opened, which prevents applications using QUIC from making further
-progress. At this stage connections can be terminated via idle timeout or explicit
-close; see {{sec-termination}}).
+they would allow to be opened by their peer. Initial limits are advertised using
+the initial_max_streams_bidi and initial_max_streams_uni transport parameters.
+As streams are opened and closed they are consumed and the cumulative total is
+incremented. Limits can be increased using the MAX_STREAMS frame but there is no
+mechanism to reduce limits. Once stream limits are reached, no more streams can
+be opened, which prevents applications using QUIC from making further progress.
+At this stage connections can be terminated via idle timeout or explicit close;
+see {{sec-termination}}).
 
 An application that uses QUIC might communicate a cumulative stream limit but
 require the connection to be closed before the limit is reached. For example,
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -647,8 +647,8 @@ timeout values may free-up resources more quickly.
 
 Application data exchanged on streams or in datagrams defers the QUIC idle
 timeout. Applications that provide their own keep-alive mechanisms will
-therefore keep a QUIC connection alive. Applications that don't provide their
-own keep-alive might be able to use transport-layer mechanisms (see {{Section
+therefore keep a QUIC connection alive. Applications that do not provide their
+own keep-alive can use transport-layer mechanisms (see {{Section
 10.1.2 of QUIC}}, and {{resumption-v-keepalive}}). However, QUIC implementation
 interfaces for controlling such transport behavior can vary, affecting the
 robustness of such approaches.
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -465,9 +465,9 @@ streams, abrupt closure of actively used streams may be undesireable or
 detrimental. In contrast, waiting for an endpoint to exhaust the advertised
 limit may not suit application or operational needs. Applications using QUIC can
 use conservative stream limits and run to completion before enacting an
-immediate close. Alternatively, a graceful close mechanism can be used to
-communicate the intention to explicitly close the connection at some future
-point.
+immediate close. Alternatively, an application-layer graceful close mechanism
+can be used to communicate the intention to explicitly close the connection at
+some future point.
 
 # Packetization and Latency
 
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -461,29 +461,32 @@ they would allow to be opened by their peer. Initial limits are advertised
 using the initial_max_streams_bidi and initial_max_streams_uni transport
 parameters. As streams are opened and closed they are consumed and the cumulative
 total is incremented. Limits can be increased using the MAX_STREAMS frame but
-there is no mechanism to reduce limits. 
+there is no mechanism to reduce limits. Once stream limits are reached, no more
+streams can be opened, which prevents applications using QUIC from making further
+progress. At this stage connections can be terminated via idle timeout or explicit
+close; see {{sec-termination}}).
 
-An application that uses QUIC might communicate a cumulative stream limit
-but require the connection to be closed before the limit is reached. For example,
+An application that uses QUIC might communicate a cumulative stream limit but
+require the connection to be closed before the limit is reached. For example,
 stopping the server to perform scheduled maintenance. Immediate connection close
-(see {{connection-termination}}) causes abrupt closure of actively used streams.
-Depending on how an application uses QUIC streams, this could be undesireable
-or detrimental to behaviour or performance. An alternative to immediate close
-is to wait for a peer to consume all of the advertised stream limit. However, the
-period of time it takes to do so is dependent on the client and an unpredictable
-closing period might not fit application or operational needs. Applications using
-QUIC can be conservative with open stream limits in order to reduce the commitment
-and indeterminism. However, being overly conservative with stream limits affects
-stream concurrency. Balancing these aspects can be specific to applications and
-their deployments. Instead of relying on stream limits to avoid abrupt closure,
-an application-layer graceful close mechanism can be used to communicate the
-intention to explicitly close the connection at some future point. HTTP/3 provides
-such a mechanism using the GOWAWAY frame. In HTTP/3,
-when the GOAWAY frame is received by a client, it
+causes abrupt closure of actively used streams. Depending on how an application
+uses QUIC streams, this could be undesirable or detrimental to behavior or
+performance. A more graceful closure technique is to stop sending increases to
+stream limits and allow the connection to naturally terminate once remaining
+streams are consumed. However, the period of time it takes to do so is dependent
+on the client and an unpredictable closing period might not fit application or
+operational needs. Applications using QUIC can be conservative with open stream
+limits in order to reduce the commitment and indeterminism. However, being
+overly conservative with stream limits affects stream concurrency. Balancing
+these aspects can be specific to applications and their deployments. Instead of
+relying on stream limits to avoid abrupt closure, an application-layer graceful
+close mechanism can be used to communicate the intention to explicitly close the
+connection at some future point. HTTP/3 provides such a mechanism using the
+GOWAWAY frame. In HTTP/3, when the GOAWAY frame is received by a client, it
 stops opening new streams even if the cumulative stream limit would allow.
 Instead the client would create a new connection on which to open further
-streams.  Once all streams are closed on the old connection, it can be terminated
-safely be a connection close or after expiration of the idle time out
+streams.  Once all streams are closed on the old connection, it can be
+terminated safely by a connection close or after expiration of the idle time out
 (see also {{sec-termination}}).
 
 # Packetization and Latency
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -478,10 +478,13 @@ stream concurrency. Balancing these aspects can be specific to applications and
 their deployments. Instead of relying on stream limits to avoid abrupt closure,
 an application-layer graceful close mechanism can be used to communicate the
 intention to explicitly close the connection at some future point. HTTP/3 provides
-such a mechanism using the GOWAWAY frame. When received by a client, it
-stops opening new streams even if the cumulative stream limit would allow. The 
-client can create a new connection on which to open further streams, once all
-streams close on the old connection it can be closed safely.
+such a mechanism using the GOWAWAY frame. In HTTP/3,
+when the GOAWAY frame is received by a client, it
+stops opening new streams even if the cumulative stream limit would allow.
+Instead the client would create a new connection on which to open further
+streams.  Once all streams are closed on the old connection, it can be terminated
+safely be a connection close or after expiration of the idle time out
+(see also {{sec-termination}}).
 
 # Packetization and Latency
 
@@ -621,7 +624,7 @@ to the server instance. The server can provide an IPv4 and an IPv6 address in a
 transport parameter during the TLS handshake and the client can select between
 the two if both are provided. See also {{Section 9.6 of QUIC}}.
 
-# Connection Termination
+# Connection Termination {#sec-termination}
 
 QUIC connections are terminated in one of three ways: implicit idle timeout,
 explicit immediate close, or explicit stateless reset.
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -503,7 +503,7 @@ is permitted, including generic ones.
 Applications using QUIC define an error
 code space that is independent from QUIC or other applications (see, for
 example, {{Section 8.1 of QUIC-HTTP}}). The values in an application error code
-space are reused across connection-level and stream-level errors.
+space can be reused across connection-level and stream-level errors.
 
 Connection errors lead to connection termination. They are signaled using a
 CONNECTION_CLOSE frame, which contains an error code and a reason field that can
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -645,14 +645,13 @@ connection. An application may adjust the maximum idle timeout for new
 connections based on the number of open or expected connections, since shorter
 timeout values may free-up resources more quickly.
 
-As long as application data is exchanged, the idle timeout does not expire. If
-an application has no data to send but desires to keep the connection open for
-longer than the timeout period, it can send keep-alive messages. A QUIC
-implementation may provide an option to defer the time-out by sending keep-alive
-messages at the transport layer to avoid unnecessary load, as specified in
-{{Section 10.1.2 of QUIC}}. Alternatively, applications using QUIC could define
-their own mechanism, such as an application-layer ping, that achieves a similar
-result. See {{resumption-v-keepalive}} for further guidance on keep-alives.
+Application data exchanged on streams or in datagrams defers the QUIC idle
+timeout. Applications that provide their own keep-alive mechanisms will
+therefore keep a QUIC connection alive. Applications that don't provide their
+own keep-alive might be able to use transport-layer mechanisms (see {{Section
+10.1.2 of QUIC}}, and {{resumption-v-keepalive}}). However, QUIC implementation
+interfaces for controlling such transport behavior can vary, affecting the
+robustness of such approaches.
 
 An immediate close is signaled by a CONNECTION_CLOSE frame (see
 {{error-handling}}). Immediate close causes all streams to become immediately
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -456,18 +456,32 @@ being terminated in some protocols.
 
 # Stream Limit Commitments
 
-QUIC endpoints can manage the cumulative maximum number of streams they would
-allow to be opened using the MAX_STREAMS frame, but there is no mechanism to
-reduce the value. An application that uses QUIC might commit to a number of
-openable streams but require the connection to be closed (for example, a
-scheduled maintenance period). Depending on how an application uses QUIC
-streams, abrupt closure of actively used streams may be undesireable or
-detrimental. In contrast, waiting for an endpoint to exhaust the advertised
-limit may not suit application or operational needs. Applications using QUIC can
-use conservative stream limits and run to completion before enacting an
-immediate close. Alternatively, an application-layer graceful close mechanism
-can be used to communicate the intention to explicitly close the connection at
-some future point.
+QUIC endpoints are responsible for communicating the cumulative limit of streams
+they would allow to be opened by their peer. Initial limits are advertised
+using the initial_max_streams_bidi and initial_max_streams_uni transport
+parameters. As streams are opened and closed they are consumed and the cumulative
+total is incremented. Limits can be increased using the MAX_STREAMS frame but
+there is no mechanism to reduce limits. 
+
+An application that uses QUIC might communicate a cumulative stream limit
+but require the connection to be closed before the limit is reached. For example,
+stopping the server to perform scheduled maintenance. Immediate connection close
+(see {{connection-termination}}) causes abrupt closure of actively used streams.
+Depending on how an application uses QUIC streams, this could be undesireable
+or detrimental to behaviour or performance. An alternative to immediate close
+is to wait for a peer to consume all of the advertised stream limit. However, the
+period of time it takes to do so is dependent on the client and an unpredictable
+closing period might not fit application or operational needs. Applications using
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-applicability.md
```diff
@@ -487,6 +487,30 @@ Padding can also be used by an application to reduce leakage of
 information about the data that is sent. A QUIC implementation can expose an
 interface that allows an application layer to specify how to apply padding.
 
+# Error Handling
+
+QUIC recommends that endpoints that detect errors should signal the occurance to
```
martinthomson, line 492:
```suggestion
QUIC recommends that endpoints that detect errors signal the occurance to
```

draft-ietf-quic-applicability.md
```diff
@@ -487,6 +487,30 @@ Padding can also be used by an application to reduce leakage of
...
+the peer. Errors can occur at the transport level and the application level.
+Transport errors, such as a protocol violation, affect the entire connection.
+Applications that use QUIC can define their own error detection and signalling
+(see, for example, {{Section 8 of QUIC-HTTP}}). Application errors can affect an
+entire connection or a single stream.
+
+QUIC defines an error code space that is used for error handling. QUIC
```
martinthomson, line 499:
```suggestion
QUIC defines an error code space that is used for error handling at the transport layer. QUIC
```

draft-ietf-quic-applicability.md
```diff
@@ -487,6 +487,30 @@ Padding can also be used by an application to reduce leakage of
...
+encourages endpoints to use the most-specific code, although any applicable code
+is permitted including generic ones. Applications using QUIC can define an error
```
martinthomson, line 501:
```suggestion
is permitted including generic ones.

Applications using QUIC can define an error
```

draft-ietf-quic-applicability.md
```diff
@@ -487,6 +487,30 @@ Padding can also be used by an application to reduce leakage of
...
+code space that is independent from QUIC or other applications (see, for
+example, {{Section 8.1 of QUIC-HTTP}}). The values in an application error code
+space are reused across connection-level and stream-level errors.
```
martinthomson, line 504:
```suggestion
example, {{Section 8.1 of QUIC-HTTP}}). The values in an application error code
space can be reused across connection-level and stream-level errors.
```

This is not completely necessary, though this is what HTTP does.  They could be separate spaces if that was what people wanted.  That is, if 2^62 error codes is not enough, you can at least double the space.

draft-ietf-quic-applicability.md
```diff
@@ -487,6 +487,30 @@ Padding can also be used by an application to reduce leakage of
...
+
+Connection errors lead to connection termination. They are signaled using a
+CONNECTION_CLOSE frame, which contains an error code and a reason field that can
+be zero length. CONNECTION_CLOSE has two frame type values: 0x1c is used for
+QUIC-level errors, 0x1d is used for application-level errors.
```
martinthomson, line 509:
```suggestion
be zero length. Different types of CONNECTION_CLOSE frame are used to
signal transport and application errors.
```

draft-ietf-quic-applicability.md
```diff
@@ -566,42 +590,49 @@ to the server instance. The server can provide an IPv4 and an IPv6 address in a
...
+explicit immediate close, or explicit stateless reset.
+
+QUIC idle timeout is enabled via transport parameters. Client and server
+announce a timeout period and the effective value for the connection is the
+minimum of the two values. After the timeout period elapses, the connection is
+silently closed. An application therefore should be able to configure its own
+maximum value, as well as have access to the computed minimum value for this
+connection. An application may adjust the maximum idle timeout for new
+connections based on the number of open or expected connections, since shorter
+timeout values may free-up memory more quickly.
```
martinthomson, line 605:
```suggestion
timeout values may free-up resources more quickly.
```

draft-ietf-quic-applicability.md
```diff
@@ -566,42 +590,49 @@ to the server instance. The server can provide an IPv4 and an IPv6 address in a
...
+
+An immediate close is signalled by a CONNECTION_CLOSE frame (see
+{{error-handling}}). Immediate close causes all streams to become immediately
+closed. QUIC endpoints can manage the cumulative maximum number of streams they
+would allow to be opened using the MAX_STREAMS frames but there is no mechanism
+to reduce the value. An application that uses QUIC might commit to a number of
+openable streams but require the connection to be closed (for example, a
+scheduled maintenance period). Depending on how an application uses QUIC streams
+(see {{use-of-streams}}), abrupt closure of actively-used streams may be
+undesireable or detrimental. In contrast, waiting for an endpoint to exhaust the
```
martinthomson, line 624: This bit about streams completely surprised me.  It's now into flow control and maximum stream limits, which really don't belong here.  If you like the text, I would suggest finding it another home.

draft-ietf-quic-applicability.md
```diff
@@ -566,42 +590,49 @@ to the server instance. The server can provide an IPv4 and an IPv6 address in a
...
+advertised limit may not suit application or operational needs. Applications
+using QUIC can use conservative stream limits and run to completion before
+enacting and immediate close. Alternatively, a graceful close mechanim can be
+used to commicate the intention to explicitly close the connection at some
+future point. QUIC does not provide any mechanism for graceful connection
+termination, applications using QUIC can define their own graceful termination
+process (see, for example, {{Section 5.2 of QUIC-HTTP}}).
```
martinthomson, line 631: The text on graceful close should be its own paragraph.

draft-ietf-quic-applicability.md
```diff
@@ -566,42 +590,49 @@ to the server instance. The server can provide an IPv4 and an IPv6 address in a
...
+
+A stateless reset is an option of last resort for an endpoint that does not have
+access to connection state. It is not expected that application using QUIC need
+information or knowledge that a stateless reset was triggered.
```
martinthomson, line 635: This is a little odd.  I would phrase the last sentence here more along the lines of "receiving a stateless reset is an indication of an unrecoverable error distinct from connection errors in that there is no application-layer information provided"

martinthomson (commented):
This looked good up until the point where it started to talk about stream limit management.

draft-ietf-quic-applicability.md
mirjak: Yes, section 4 is on use of streams. Maybe we can just add another short subsection there to talk about the max stream limit.

draft-ietf-quic-applicability.md
MikeBishop: Or perhaps 
```suggestion
QUIC recommends that endpoints signal any detected errors to
```
...as both more concise and removes the mis-spelling of "occurrence."

draft-ietf-quic-applicability.md
```diff
@@ -487,6 +487,30 @@ Padding can also be used by an application to reduce leakage of
...
+
+QUIC recommends that endpoints that detect errors should signal the occurance to
+the peer. Errors can occur at the transport level and the application level.
+Transport errors, such as a protocol violation, affect the entire connection.
+Applications that use QUIC can define their own error detection and signalling
+(see, for example, {{Section 8 of QUIC-HTTP}}). Application errors can affect an
+entire connection or a single stream.
+
+QUIC defines an error code space that is used for error handling. QUIC
+encourages endpoints to use the most-specific code, although any applicable code
```
MikeBishop, line 500:
```suggestion
encourages endpoints to use the most specific code, although any applicable code
```

draft-ietf-quic-applicability.md
MikeBishop, line 501:
```suggestion
is permitted, including generic ones.

Applications using QUIC define an error
```
It's not a "can," since the spaces are distinct.  The application defines any semantics that might exist.

draft-ietf-quic-applicability.md
```diff
@@ -566,42 +590,49 @@ to the server instance. The server can provide an IPv4 and an IPv6 address in a
...
+implementation may provide an option to defer the time-out by sending keep-alive
+messages at the transport layer to avoid unnecessary load, as specified in
+{{Section 10.1.2 of QUIC}}. Alternatively, applications using QUIC could define
+their own mechanism, such as an application-layer ping, that achieves a similar
+result. See {{resumption-v-keepalive}} for further guidance on keep-alives.
+
+An immediate close is signalled by a CONNECTION_CLOSE frame (see
+{{error-handling}}). Immediate close causes all streams to become immediately
+closed. QUIC endpoints can manage the cumulative maximum number of streams they
+would allow to be opened using the MAX_STREAMS frames but there is no mechanism
```
MikeBishop, line 619:
```suggestion
would allow to be opened using the MAX_STREAMS frame, but there is no mechanism
```

draft-ietf-quic-applicability.md
```diff
@@ -566,42 +590,49 @@ to the server instance. The server can provide an IPv4 and an IPv6 address in a
...
+to reduce the value. An application that uses QUIC might commit to a number of
+openable streams but require the connection to be closed (for example, a
+scheduled maintenance period). Depending on how an application uses QUIC streams
+(see {{use-of-streams}}), abrupt closure of actively-used streams may be
```
MikeBishop, line 623:
```suggestion
(see {{use-of-streams}}), abrupt closure of actively used streams may be
```

draft-ietf-quic-applicability.md
```diff
@@ -566,42 +590,49 @@ to the server instance. The server can provide an IPv4 and an IPv6 address in a
...
+undesireable or detrimental. In contrast, waiting for an endpoint to exhaust the
+advertised limit may not suit application or operational needs. Applications
+using QUIC can use conservative stream limits and run to completion before
+enacting and immediate close. Alternatively, a graceful close mechanim can be
```
MikeBishop, line 627:
```suggestion
enacting an immediate close. Alternatively, a graceful close mechanism can be
```

draft-ietf-quic-applicability.md
```diff
@@ -566,42 +590,49 @@ to the server instance. The server can provide an IPv4 and an IPv6 address in a
...
+used to commicate the intention to explicitly close the connection at some
```
MikeBishop, line 628:
```suggestion
used to communicate the intention to explicitly close the connection at some
```

draft-ietf-quic-applicability.md
```diff
@@ -566,42 +590,49 @@ to the server instance. The server can provide an IPv4 and an IPv6 address in a
...
+future point. QUIC does not provide any mechanism for graceful connection
+termination, applications using QUIC can define their own graceful termination
```
MikeBishop, line 630:
```suggestion
termination; applications using QUIC can define their own graceful termination
```

MikeBishop (commented):
I like the general outline, but lots of editorial nits here.  Also, you're inconsistent on whether the "l" gets doubled in "signaling/signalling" and "signaled/signalled."

draft-ietf-quic-applicability.md
LPardue: I like Mike's

draft-ietf-quic-applicability.md
LPardue: I think we actually need to go one step further. My original intent was to say "applications can (and probably should define some codes". But with the comments and suggestions in this PR, I think the clearer statement is more like

```
Applications using QUIC are responsible for defining the error codes used to signal connection-level or stream-level errors (see, for example, {{Section 8.1 of QUIC-HTTP}}). Error code values are independent from QUIC or other applications. Values can be shared between the error types.
```

draft-ietf-quic-applicability.md
LPardue: Agree, although this did kind of blindside me. I think QUIC transport's term "application protocol error code space" really reinforced my incorrect mental model.

draft-ietf-quic-applicability.md
LPardue: Will see what I can do. This text was really the main thing I was looking for in my issue #175 because shutdowns affect streams but QUIC doesn't talk about what termination of streams means for an app. I realise now that a subsection for stream limits and termination is probably useful, and then this part can cross-referencet to that as a "global reset of all streams" type thing

draft-ietf-quic-applicability.md
MikeBishop: I'd be okay with that text.

LPardue:
Many thanks for the editorial feedback. I believe this is now addresses it all. I'll defer to @mirjak to merge this one, I strongly suggest the "squash and merge button" :)

draft-ietf-quic-applicability.md
```diff
@@ -566,42 +607,41 @@ to the server instance. The server can provide an IPv4 and an IPv6 address in a
...
+timeout values may free-up resources more quickly.
+
+As long as application data is exchanged, the idle timeout does not expire. If
+an application has no data to send but desires to keep the connection open for
+longer than the timeout period, it can send keep-alive messages. A QUIC
+implementation may provide an option to defer the time-out by sending keep-alive
+messages at the transport layer to avoid unnecessary load, as specified in
+{{Section 10.1.2 of QUIC}}. Alternatively, applications using QUIC could define
+their own mechanism, such as an application-layer ping, that achieves a similar
+result. See {{resumption-v-keepalive}} for further guidance on keep-alives.
```
mirjak, line 635: I'd say you "only" should use application layer ping if you also need that information for the application layer logic itself. If you do application layer pings that just looks like payload traffic on the transport which means even if QUIC is configured to send keep-alives it will not send any because it never becomes idle. So not sure if we actually need to say anything about application layer pings here at all...?

draft-ietf-quic-applicability.md
```diff
@@ -454,6 +454,21 @@ Some deadlocking scenarios might be resolved by cancelling affected streams with
...
+# Stream Limit Commitments
+
+QUIC endpoints can manage the cumulative maximum number of streams they would
+allow to be opened using the MAX_STREAMS frame, but there is no mechanism to
+reduce the value. An application that uses QUIC might commit to a number of
+openable streams but require the connection to be closed (for example, a
+scheduled maintenance period). Depending on how an application uses QUIC
+streams, abrupt closure of actively used streams may be undesireable or
+detrimental. In contrast, waiting for an endpoint to exhaust the advertised
+limit may not suit application or operational needs. Applications using QUIC can
```
mirjak, line 466: I think I miss some connection here: why should an endpoint ever wait until all stream are exhausted...?

draft-ietf-quic-applicability.md
```diff
@@ -454,6 +454,21 @@ Some deadlocking scenarios might be resolved by cancelling affected streams with
...
+use conservative stream limits and run to completion before enacting an
+immediate close. Alternatively, a graceful close mechanism can be used to
```
mirjak, line 468: Do you mean graceful close at the application layer?

draft-ietf-quic-applicability.md
LPardue: So imagine an HTTP/3 server that advertises that up to 10 bidirectional client streams streams can be created. The client has created 2 streams for requests,and the responses are large payloads. 

The server gets scheduled for an upgrade and the operator wants to drain the connection gracefully. It is happy to let the large downloads conclude but it wants new requests to get routed to a new machine. Any form of connection closure at this stage risks affecting the in-flight response. Connection closure when all streams are exhausted is a much safer option.

The client believes it can create 8 more streams and it is sub optimal to create a new connection while the current one appears to be able to handle what is needed. It could also be bad to open concurrent connections opportunistically in anticipation of a server having to gracefully close. The server could just wait for the client to create those 8 streams and the instantly reject them with STOP_SENDING, RESET_STREAM, or an HTTP response. That's a very chatty process, and might delay the client from being able to fulfill its duties in a timely manner. 

HTTP/3 GOAWAY is a graceful close signal that permits server to tell the client that it won't accept new requests, but it doesn't interrupt active streams. A client that receives GOAWAY could decide to open a new connection in parallel to the active transfers completing.

draft-ietf-quic-applicability.md
LPardue: ah good spot. I lost the context when moving the text to this section, so let me make it clear again.

draft-ietf-quic-applicability.md
mirjak: I understand why you need/want GOAWAY at the application.I'm just think that waiting until stream are exhausted is never really an option because as soon as one stream is closed the client can immediately open another one. I think the two options you have is you either wait until the connection is done entirely, or you have an application layer mechanism that indicates that no new streams should be created.

LPardue: The client can only open as many as MAX_STREAMs. So if a server wants to drain a connection without a GOWAY, it can stop granting stream credits (this is what I mean by exhaustion) and eventually the client will run out of things to do. From the other guidance in the spec, a client with nothing to do should eventually give up either via idle timeout or explicit close.

mirjak: Okay, you are right. I was misremembering how MAX_STREAM works (would maybe make sense to add a sentence). So basically what you are saying is that if the server wants to close the connection but doesn't want to interrupt on-going activities, the only thing it can do is to not issue any new stream credits and wait until the client is done because it easier has nothing to send anymore or reaches the stream limit (if you don't have an graceful close on application level). I think this can be worded simpler than taking about "exhausting streams" :-)

LPardue: I updated the text to address your comments,  PTAL

draft-ietf-quic-applicability.md
```diff
@@ -454,6 +454,35 @@ Some deadlocking scenarios might be resolved by cancelling affected streams with
...
+total is incremented. Limits can be increased using the MAX_STREAMS frame but
+there is no mechanism to reduce limits. 
+
+An application that uses QUIC might communicate a cumulative stream limit
+but require the connection to be closed before the limit is reached. For example,
+stopping the server to perform scheduled maintenance. Immediate connection close
+(see {{connection-termination}}) causes abrupt closure of actively used streams.
+Depending on how an application uses QUIC streams, this could be undesireable
+or detrimental to behaviour or performance. An alternative to immediate close
+is to wait for a peer to consume all of the advertised stream limit. However, the
```
mirjak, line 472:
```suggestion
is to not issue any new stream credits and wait for a peer to close the connection,
either because there is no additional data to send or it is forced to close when it
consumed all of the advertised stream limit. However, the
```
Getting a bit lengthly; it's just a proposal...

draft-ietf-quic-applicability.md
```diff
@@ -454,6 +454,35 @@ Some deadlocking scenarios might be resolved by cancelling affected streams with
...
+period of time it takes to do so is dependent on the client and an unpredictable
+closing period might not fit application or operational needs. Applications using
+QUIC can be conservative with open stream limits in order to reduce the commitment
+and indeterminism. However, being overly conservative with stream limits affects
+stream concurrency. Balancing these aspects can be specific to applications and
+their deployments. Instead of relying on stream limits to avoid abrupt closure,
+an application-layer graceful close mechanism can be used to communicate the
+intention to explicitly close the connection at some future point. HTTP/3 provides
+such a mechanism using the GOWAWAY frame. When received by a client, it
```
mirjak, line 481:
```suggestion
such a mechanism using the GOWAWAY frame. In HTTP/3,
when the GOAWAY frame is received by a client, it
```

draft-ietf-quic-applicability.md
```diff
@@ -454,6 +454,35 @@ Some deadlocking scenarios might be resolved by cancelling affected streams with
...
+stops opening new streams even if the cumulative stream limit would allow. The 
+client can create a new connection on which to open further streams, once all
+streams close on the old connection it can be closed safely.
```
mirjak, line 484:
```suggestion
stops opening new streams even if the cumulative stream limit would allow.
Instead the client would create a new connection on which to open further
streams.  Once all streams are closed on the old connection, it can be terminated
safely be a connection close or after expiration of the idle time out
(see also {{sec-termination}}).
```

draft-ietf-quic-applicability.md
```diff
@@ -566,42 +621,41 @@ to the server instance. The server can provide an IPv4 and an IPv6 address in a
...
-adjust the maximum idle timeout for new connections based on the number of open
-or expected connections, since shorter timeout values may free-up memory more
-quickly.
-
-If an application desires to keep the connection open for longer than the
-announced timeout, it can send keep-alive messages; a QUIC implementation may
-provide an option to defer the time-out by sending keep-alive messages at the
-transport layer to avoid unnecessary load, as specified in {{Section 10.1.2 of
-QUIC}}. See {{resumption-v-keepalive}} for further guidance on keep-alives.
+# Connection Termination
```
mirjak, line 624:
```suggestion
# Connection Termination {#sec-termination}
```

mirjak (commented):
proposed edits...

draft-ietf-quic-applicability.md
LPardue: how about we put your suggested text about how endpoints might close when the limit is reached, then the sentence here can say 

"An alternative to immediate close is to stop increasing the stream limit and wait for the peer to consume the remaining streams."

?

mirjak: Sounds good!

draft-ietf-quic-applicability.md
LPardue: That's already what this paragraph says IMO - if there is no useful application data to send, you can use QUIC keepalives or design an application mechanism. The important thing is that applications might have their own idle timeouts above QUIC (for example, time between HTTP requests, time between DATA frames) and its totally fine for applications to have those and design mechanisms that avoid them being triggered.

draft-ietf-quic-applicability.md
```diff
@@ -454,6 +454,41 @@ Some deadlocking scenarios might be resolved by cancelling affected streams with
...
+parameters. As streams are opened and closed they are consumed and the cumulative
+total is incremented. Limits can be increased using the MAX_STREAMS frame but
+there is no mechanism to reduce limits. Once stream limits are reached, no more
+streams can be opened, which prevents applications using QUIC from making further
+progress. At this stage connections can be terminated via idle timeout or explicit
+close; see {{sec-termination}}).
+
+An application that uses QUIC might communicate a cumulative stream limit but
+require the connection to be closed before the limit is reached. For example,
+stopping the server to perform scheduled maintenance. Immediate connection close
```
mirjak, line 471:
```suggestion
to stop the server to perform scheduled maintenance. Immediate connection close
```

draft-ietf-quic-applicability.md
mirjak: Yes, the tiny different I would be looking for is to say it more like: if there is no useful application data to send, you should use QUIC keepalives or only design an application mechanism if that is needed on the app layer anyway for something else. Or to put if differently: please don't design your own app layer ping just because you can.

draft-ietf-quic-applicability.md
```diff
@@ -454,6 +454,41 @@ Some deadlocking scenarios might be resolved by cancelling affected streams with
...
+these aspects can be specific to applications and their deployments. Instead of
+relying on stream limits to avoid abrupt closure, an application-layer graceful
+close mechanism can be used to communicate the intention to explicitly close the
+connection at some future point. HTTP/3 provides such a mechanism using the
+GOWAWAY frame. In HTTP/3, when the GOAWAY frame is received by a client, it
+stops opening new streams even if the cumulative stream limit would allow.
+Instead the client would create a new connection on which to open further
+streams.  Once all streams are closed on the old connection, it can be
+terminated safely by a connection close or after expiration of the idle time out
+(see also {{sec-termination}}).
```
mirjak, line 490: okay this reference to the other section is now a bit redundant with the last sentence of the previous paragraph but might not hurt...

draft-ietf-quic-applicability.md
LPardue: I disagree. An application protocol designer probably wants wide interoperability and they could be sorely disappointed to find that QUIC implementations don't all  expose a keep alive mechanism. The best way to mitigate that risk is to pretend transport keepalives don't exist and to always have an application one. Doing it that way, means that all implementations of an application protocol can expect the same behaviour.

The text here attempts to tread a fine line by stating what is possible but without recommending any particular course of action.

mirjak: You are right on the point that this discussion actually doesn't belong in this document as that question is not QUIC specific. Therefore I would still advocate for removing the second to last sentence here about app layer pings because of course applications can always do that but is also not specific for applications that use QUIC.

LPardue: Hmm. So given that this doc is focussed on pointing out caveats for applications, how about we swizzle the text a bit to say 

"Application data exchanged on streams or in datagrams defers the QUIC idle timeout. Applications that provide their own keep-alive mechanisms will therefore keep a QUIC connection alive. Applications that don't provide their own keep-alive might be able to use transport-layer mechanism (see {{Section 10.1.2 of QUIC}}, and {{resumption-v-keepalive}}). However, QUIC implementation interfaces for controlling such transport behaviour can vary, affecting the robustness of such designs.

mirjak: WFM. Thanks!

draft-ietf-quic-applicability.md
```diff
@@ -566,42 +627,40 @@ to the server instance. The server can provide an IPv4 and an IPv6 address in a
...
+silently closed. An application therefore should be able to configure its own
+maximum value, as well as have access to the computed minimum value for this
+connection. An application may adjust the maximum idle timeout for new
+connections based on the number of open or expected connections, since shorter
+timeout values may free-up resources more quickly.
+
+Application data exchanged on streams or in datagrams defers the QUIC idle
+timeout. Applications that provide their own keep-alive mechanisms will
+therefore keep a QUIC connection alive. Applications that don't provide their
+own keep-alive might be able to use transport-layer mechanisms (see {{Section
```
mirjak, line 651:
```suggestion
therefore keep a QUIC connection alive. Applications that do not provide their
own keep-alive can use transport-layer mechanisms (see {{Section
```

PR #299, opened 2021-03-24
Try to make flow control consuming and processing crisper

### LINKED ISSUE #193: App: Sect 4.3. Flow Control - we should have a subsection also encouraging timely credits
Issue: Sect 4.3. Flow Control - we should have a subsection also encouraging timely credits

This ID discusses flow control deadlock, whichI think  is important. I suggest we should also note that timely release of flow control credit is also important (QUIC transport has some text to point to). This need is evident when the path has a large RTT, and there can be a large BDP, where the whole of operation of QUIC can be dominated by the way in which the application manages the flow credit. This might not be obvious, because the equivalent rwnd update in TCP is handled by the transport rather than the application.

LPardue:
Fixes #193.

This mainly wiggles the existing text around and tries to fully qualify transport vs application to help explain the ways an implementation might do things, which is independent of how an application protocol might be designed (for example H3 doesn't require incremental processing but most people come to a realizatopm that it is the best way to do things).

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -425,7 +425,8 @@ transport-layer and application-layer receive buffers. Consuming data does not
 always imply it is immediately processed. However, a common flow control
 implementation technique is to extend credit to the sender, by emitting MAX_DATA
 and/or MAX_STREAM_DATA frames, as data is consumed. Delivery of these frames
-is affected by uplink latency. If credit is not extended in a timely manner, the
+is affected by the latency of the back channel from the receiver to the data sender.
+If credit is not extended in a timely manner, the
 sending application can be blocked, effectively throttling the downlink.
 
 Large application messages can produce deadlocking if the recipient does not
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -418,9 +418,9 @@ whether they become a problem depends on how implementations consume data and
 provide flow control credit.  Understanding what causes deadlocking might help
 implementations avoid deadlocks.
 
+Timely updates of transport flow control credit can improve performance.
 Applications that use QUIC often have a data consumer that reads data from
-transport buffers. Timely updates of transport flow control credit can improve
-performance. A common flow control implementation technique is for a QUIC
+transport buffers. A common flow control implementation technique is for a QUIC
 receiver to extend credit to the sender as the data consumer reads data.  Some
 implementations might have independent transport-layer and application-layer
 receive buffers. Consuming data does not always imply it is immediately
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -407,7 +407,7 @@ change with its peer might encounter performance issues or deadlocks.
 
 ## Flow Control Deadlocks {#flow-control-deadlocks}
 
-Flow control provides a means of managing access to the limited buffers
+QUIC flow control provides a means of managing access to the limited buffers
 endpoints have for incoming data.  This mechanism limits the amount of data that
 can be in buffers in endpoints or in transit on the network.  However, there are
 several ways in which limits can produce conditions that can cause a connection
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-applicability.md
```diff
@@ -418,28 +418,34 @@ whether they become a problem depends on how implementations consume data and
...
-connection flow control limits can be consumed by other streams.
-
-A common flow control implementation technique is for a receiver to extend
-credit to the sender as a the data consumer reads data. In this setting, a
-length-prefixed message format makes it easier for the data consumer to leave
-data unread in the receiver's buffers and thereby withhold flow control credit.
-If flow control limits prevent the remainder of a message from being sent, a
+Applications that use QUIC often have a data consumer that reads data from
+transport buffers. Timely updates of transport flow control credit can improve
+performance. A common flow control implementation technique is for a QUIC
```
martinthomson, line 423: Reorder these first two sentences:

```suggestion
Timely updates of transport flow control credit can improve performance.
Applications that use QUIC often have a data consumer that reads data from
transport buffers. A common flow control implementation technique is for a QUIC
```

draft-ietf-quic-applicability.md
```diff
@@ -418,28 +418,34 @@ whether they become a problem depends on how implementations consume data and
...
+receiver to extend credit to the sender as the data consumer reads data.  Some
+implementations might have independent transport-layer and application-layer
+receive buffers. Consuming data does not always imply it is immediately
+processed. However, when data is consumed,
```
martinthomson, line 427: Incomplete sentence.

gorryfair:
The new text is good and clear, but it doesn't really explain that the when credit is released at the receiver, could we say a little more: it takes 1/2 RTT for cerdit to reach the sender, that causes new data to be released, for a large RTT, this means that credit has to be sufficient to avoid blocking or to be sent promptly, otherwise transfers will stall for periods of an RTT. (I know TCP does this stuff for the app and grows the window, but QUIC I suspect needs something to happen to release the credit). I know people closer to the Apps can write this sentence better than me; from my perspective I just see the effects when an app blocks or release credit in big chunks... over a large path RTT and the network performance suffers.

LPardue:
Thanks Gorry. I think I agree. How about I tweak "Timely updates can improve performance" paragraph to say "Timing of updates can affect performance". I can add a mention that the extension of credits on the data recever side is enacted by emitting MAX_DATA/MAX_STREAM_DATA frames, which take at least a 1/2 RTT to reach the data sender side.

I think the useful additional applicability statement to be made based on the above is that the data consumer read rate may be disjoint from the delivery rate and responsiveness of signalling channel back. For a highly asymmetric link, the download rate can be application-limited well below the theoretical bandwidth simply because an application is spending too much share of the time reading packets and not giving enough time to emit QUIC packets. 

Does that sound ok, or am I missing your ask?

gorryfair:
This is definitely heading in the direction I was hoping, it helps avoid people developing on a LAN and then discovering the effects on other paths later. Please add something.

mirjak (changes requested):
This PR is fine, only that there is an incomplete sentence...

LPardue:
I've reworked the paragraph with my attempt to address @gorryfair's comment. I've introduced the terms uplink and downlink, which I'm not totally happy with. If someone has a better suggestion I'm all ears.

draft-ietf-quic-applicability.md
```diff
@@ -418,28 +418,36 @@ whether they become a problem depends on how implementations consume data and
...
-data unread in the receiver's buffers and thereby withhold flow control credit.
-If flow control limits prevent the remainder of a message from being sent, a
+The size and rate of transport flow control credit updates can affect
+performance. Applications that use QUIC often have a data consumer that reads
+data from transport buffers. Some implementations might have independent
+transport-layer and application-layer receive buffers. Consuming data does not
+always imply it is immediately processed. However, a common flow control
+implementation technique is to extend credit to the sender, by emitting MAX_DATA
+and/or MAX_STREAM_DATA frames, as data is consumed. Delivery of these frames
+is affected by uplink latency. If credit is not extended in a timely manner, the
```
mirjak, line 428:
```suggestion
is affected by the latency of the back channel from the receiver to the data sender.
If credit is not extended in a timely manner, the
```
Is this any better?

draft-ietf-quic-applicability.md
```diff
@@ -418,28 +418,37 @@ whether they become a problem depends on how implementations consume data and
...
+The size and rate of transport flow control credit updates can affect
+performance. Applications that use QUIC often have a data consumer that reads
+data from transport buffers. Some implementations might have independent
+transport-layer and application-layer receive buffers. Consuming data does not
+always imply it is immediately processed. However, a common flow control
+implementation technique is to extend credit to the sender, by emitting MAX_DATA
+and/or MAX_STREAM_DATA frames, as data is consumed. Delivery of these frames
+is affected by the latency of the back channel from the receiver to the data sender.
+If credit is not extended in a timely manner, the
+sending application can be blocked, effectively throttling the downlink.
```
britram, line 430:
```suggestion
sending application can be blocked, effectively throttling the connection.
```

mirjak (approved):

PR #284, opened 2021-03-02
Manageability editorial pass

MikeBishop:
Mostly nits, some text about versions.  Some reflow, and fixed up section references to other documents.  If there's anything you think needs to be lifted out to a separate PR/issue, I'm happy to do that.

PR #305, opened 2021-03-25
Update on UDP policing section

### LINKED ISSUE #295: Recommendation about using 4-tuple for NAT
Comment from PR #272 from @ianswett on UDP policing:

"This is recommending something that relies on the 4-tuple, but particularly if there's another NAT involved, the 4-tuple may not be stable. So this puts connections in a situation where the handshake completes, they use the connection for a bit, a NAT rebind occurs, and then the connection is blackholed.

There are defenses against this problem, but this seems like a fairly dangerous recommendation."

### LINKED ISSUE #296: Dropping of UDP traffic
Comment from @martinthomson from PR #272 on UDP Policing about this phrase:

`space that corresponds to the fraction of UDP traffic one wishes to drop.`

"The extra words don't really help here. If anything, you need to highlight the need to block only a subset of addresses rather than a subset of packets."

mirjak:
fixes #284 and fixes #296

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -919,7 +919,7 @@ fixed portion of the network capacity to UDP and blocking UDP datagrams over tha
 cap. As the portion of QUIC traffic compared to TCP is also expected to
 increase over time, limits might need to be adopted dynamically.
 
-Further, if UDP traffic is throttle, it is reecommended to rather block
+Further, if UDP traffic is throttle, it is recommended to rather block
 QUIC flows entirely than dropping packets randomly. When the handshake is
 blocked, QUIC-capable applications may failover to TCP
 (at least applications using well-known UDP ports). However, blocking a
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -930,9 +930,9 @@ should be rather realised by per-flow policing, e.g, based on the hash
 of the UDP datagram's source and destination
 addresses and blocking a portion of the hash space that corresponds to the
 fraction of UDP traffic one wishes to drop.
-While QUIC endpoints are often able to survive address changes, e.g. by NAT rebindings,
-blocking a portion of the traffic based on 5-tuple hashing increases the risk of
-blackholing an active connection when the address changes.
+While QUIC endpoints are often able to survive address changes, e.g. by NAT
+rebindings, blocking a portion of the traffic based on 5-tuple hashing increases
+the risk of black-holing an active connection when the address changes.
 
 
 ## DDoS Detection and Mitigation {#sec-ddos-dec}
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -920,7 +920,7 @@ that cap. As the portion of QUIC traffic compared to TCP is also expected to
 increase over time, using such a limit is not recommended but if done,
 limits might need to be adapted dynamically.
 
-Further, if UDP traffic is throttled, it is recommended to block individual
+Further, if UDP traffic is desired to be throttled, it is recommended to block individual
 QUIC flows entirely rather than dropping packets randomly. When the handshake is
 blocked, QUIC-capable applications may failover to TCP
 However, blocking a
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -905,26 +905,31 @@ Admitting a few packets allows the QUIC endpoint to determine that the path
 accepts QUIC. Sudden drops afterwards will result in slow and costly timeouts
 before abandoning the connection.
 
-## UDP Policing {#sec-udp-1312}
+## UDP Blocking or Throttling {#sec-udp-1312}
 
 Today, UDP is the most prevalent DDoS vector, since it is easy for compromised
 non-admin applications to send a flood of large UDP packets (while with TCP the
 attacker gets throttled by the congestion controller) or to craft reflection and
-amplification attacks. Networks should therefore be
-prepared for UDP flood attacks on ports used for QUIC traffic. One possible
-response to this threat is to police UDP traffic on the network, allocating a
-fixed portion of the network capacity to UDP and blocking UDP datagram over that
-cap.
+amplification attacks. Some networks therefore block UDP traffic.
+With increased deployment of QUIC, there is also an increased need to allow
+UDP traffic on ports used for QUIC. However, if UDP is generally enabled on these
+ports, UDP flood attacks may also used the same ports. One possible
+response to this threat is to throttle UDP traffic on the network, allocating a
+fixed portion of the network capacity to UDP and blocking UDP datagrams over that
+cap. As the portion of QUIC traffic compared to TCP is also expected to
+increase over time, limits might need to be adopted dynamically.
 
-The recommended way to police QUIC
-packets is to either drop them all or to throttle them based on the hash of the
-UDP datagram's source and destination addresses, blocking a portion of the hash
-space that corresponds to the fraction of UDP traffic one wishes to drop.
-When the handshake is blocked, QUIC-capable applications may failover to TCP
+Further, if UDP traffic is throttle, it is reecommended to rather block
+QUIC flows entirely than dropping packets randomly. When the handshake is
+blocked, QUIC-capable applications may failover to TCP
 (at least applications using well-known UDP ports). However, blocking a
 random fraction of QUIC packets across 4-tuples will allow many QUIC handshakes
 to complete, preventing a TCP failover, but the connections will suffer from
-severe packet loss.
+severe packet loss (see also {{sec-filtering}}). Per-flow throttling can be
+realized based on the hash of the UDP datagram's source and destination
+addresses and blocking a portion of the hash space that corresponds to the
+fraction of UDP traffic one wishes to drop.
+
 
 ## DDoS Detection and Mitigation {#sec-ddos-dec}
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -917,7 +917,8 @@ these ports, UDP flood attacks may also use the same ports. One possible
 response to this threat is to throttle UDP traffic on the network, allocating a
 fixed portion of the network capacity to UDP and blocking UDP datagrams over
 that cap. As the portion of QUIC traffic compared to TCP is also expected to
-increase over time, limits might need to be adopted dynamically.
+increase over time, using such a limit is not recommended but if done,
+limits might need to be adapted dynamically.
 
 Further, if UDP traffic is throttled, it is recommended to block individual
 QUIC flows entirely rather than dropping packets randomly. When the handshake is
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -919,16 +919,20 @@ fixed portion of the network capacity to UDP and blocking UDP datagrams over
 that cap. As the portion of QUIC traffic compared to TCP is also expected to
 increase over time, limits might need to be adopted dynamically.
 
-Further, if UDP traffic is throttle, it is recommended to block individual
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -905,26 +905,31 @@ Admitting a few packets allows the QUIC endpoint to determine that the path
...
+amplification attacks. Some networks therefore block UDP traffic.
+With increased deployment of QUIC, there is also an increased need to allow
+UDP traffic on ports used for QUIC. However, if UDP is generally enabled on these
+ports, UDP flood attacks may also used the same ports. One possible
+response to this threat is to throttle UDP traffic on the network, allocating a
+fixed portion of the network capacity to UDP and blocking UDP datagrams over that
+cap. As the portion of QUIC traffic compared to TCP is also expected to
+increase over time, limits might need to be adopted dynamically.
+
+Further, if UDP traffic is throttle, it is reecommended to rather block
```
mirjak, line 922:
```suggestion
Further, if UDP traffic is throttle, it is recommended to rather block
```

draft-ietf-quic-manageability.md
```diff
@@ -905,26 +905,31 @@ Admitting a few packets allows the QUIC endpoint to determine that the path
...
+QUIC flows entirely than dropping packets randomly. When the handshake is
+blocked, QUIC-capable applications may failover to TCP
 (at least applications using well-known UDP ports). However, blocking a
 random fraction of QUIC packets across 4-tuples will allow many QUIC handshakes
 to complete, preventing a TCP failover, but the connections will suffer from
-severe packet loss.
+severe packet loss (see also {{sec-filtering}}). Per-flow throttling can be
+realized based on the hash of the UDP datagram's source and destination
+addresses and blocking a portion of the hash space that corresponds to the
+fraction of UDP traffic one wishes to drop.
```
ianswett, line 931: If we're going to advise this, I think we should be clear that addresses may change for QUIC connections, and when they do, this will result in post-handshake blackholing.

ianswett (commented):
This is a tricky one.  I think it could use on-list discussion?

draft-ietf-quic-manageability.md
```diff
@@ -905,26 +905,31 @@ Admitting a few packets allows the QUIC endpoint to determine that the path
...
+increase over time, limits might need to be adopted dynamically.
+
+Further, if UDP traffic is throttle, it is recommended to rather block
+QUIC flows entirely than dropping packets randomly. When the handshake is
+blocked, QUIC-capable applications may failover to TCP
 (at least applications using well-known UDP ports). However, blocking a
 random fraction of QUIC packets across 4-tuples will allow many QUIC handshakes
 to complete, preventing a TCP failover, but the connections will suffer from
-severe packet loss.
+severe packet loss (see also {{sec-filtering}}). Per-flow throttling can be
```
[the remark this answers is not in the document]
mirjak: hm, it's blocking some flow to throttle the aggregate UDP traffic...

draft-ietf-quic-manageability.md
mirjak: It could but that is a general problem on the Internet today. If the address changes it is not guaranteed that the traffic is still routed correctly...

ianswett: True, but now we have a potentially common case where a NAT rebinds and there's a chance the flow is suddenly blackholed.

draft-ietf-quic-manageability.md
```diff
@@ -905,26 +905,31 @@ Admitting a few packets allows the QUIC endpoint to determine that the path
...
+realized based on the hash of the UDP datagram's source and destination
+addresses and blocking a portion of the hash space that corresponds to the
+fraction of UDP traffic one wishes to drop.
+
```
mirjak, line 932:
```suggestion
While QUIC endpoints are often able to survive address changes, e.g. by NAT rebindings,
blocking a portion of the traffic based on 5-tuple hashing increases the risk of
blackholing an active connection when the address changes.

```

draft-ietf-quic-manageability.md
ianswett: I think so.  I still think it's worth a wider discussion, but the trade-offs are clearer now.

draft-ietf-quic-manageability.md
```diff
@@ -905,26 +905,31 @@ Admitting a few packets allows the QUIC endpoint to determine that the path
...
+these ports, UDP flood attacks may also use the same ports. One possible
+response to this threat is to throttle UDP traffic on the network, allocating a
+fixed portion of the network capacity to UDP and blocking UDP datagrams over
+that cap. As the portion of QUIC traffic compared to TCP is also expected to
+increase over time, limits might need to be adopted dynamically.
+
+Further, if UDP traffic is throttle, it is recommended to block individual
+QUIC flows entirely rather than dropping packets randomly. When the handshake is
+blocked, QUIC-capable applications may failover to TCP
 (at least applications using well-known UDP ports). However, blocking a
```
martinthomson, line 925:
```suggestion
However, blocking a
```

Well-known ports has nothing to do with it.

draft-ietf-quic-manageability.md
```diff
@@ -905,26 +905,31 @@ Admitting a few packets allows the QUIC endpoint to determine that the path
...
 random fraction of QUIC packets across 4-tuples will allow many QUIC handshakes
 to complete, preventing a TCP failover, but the connections will suffer from
-severe packet loss.
+severe packet loss (see also {{sec-filtering}}). Per-flow throttling can be
+realized based on the hash of the UDP datagram's source and destination
+addresses and blocking a portion of the hash space that corresponds to the
```
martinthomson, line 930: This bit about hash space is unnecessarily narrow.  The overall point is sound: block a subset of flows, not a subset of packets.  Talking about hashing is a distraction.

draft-ietf-quic-manageability.md
martinthomson: > QUIC connections can be resilient to address changes, such as those caused by NAT rebinding.  However, blocking a portion of flows as identified by address tuples means that some active flows will fail when addresses change.

"blackholing" or black hole is not a useful term to use here.  I would rather you avoid it.

martinthomson (approved):
Generally speaking, this is OK.  I have a few suggestions however.

draft-ietf-quic-manageability.md
mirjak: Do you have a better term?

draft-ietf-quic-manageability.md
mirjak: I disagree. This is about guidance to network function. The more specific we can be, the more likely people do the right thing.

draft-ietf-quic-manageability.md
```diff
@@ -905,26 +905,35 @@ Admitting a few packets allows the QUIC endpoint to determine that the path
...
-When the handshake is blocked, QUIC-capable applications may failover to TCP
-(at least applications using well-known UDP ports). However, blocking a
+amplification attacks. Some networks therefore block UDP traffic.
+With increased deployment of QUIC, there is also an increased need to allow
+UDP traffic on ports used for QUIC. However, if UDP is generally enabled on
+these ports, UDP flood attacks may also use the same ports. One possible
+response to this threat is to throttle UDP traffic on the network, allocating a
+fixed portion of the network capacity to UDP and blocking UDP datagrams over
+that cap. As the portion of QUIC traffic compared to TCP is also expected to
+increase over time, limits might need to be adopted dynamically.
```
mirjak, line 920:
```suggestion
increase over time, using such a limit is not recommended but if done,
limits might need to be adopted dynamically.
```

draft-ietf-quic-manageability.md
```diff
@@ -905,26 +905,35 @@ Admitting a few packets allows the QUIC endpoint to determine that the path
...
+
+Further, if UDP traffic is throttled, it is recommended to block individual
```
mirjak, line 922:
```suggestion
Further, if UDP traffic is desired to be throttled, it is recommended to block individual
```

draft-ietf-quic-manageability.md
```diff
@@ -905,26 +905,35 @@ Admitting a few packets allows the QUIC endpoint to determine that the path
...
 to complete, preventing a TCP failover, but the connections will suffer from
-severe packet loss.
+severe packet loss (see also {{sec-filtering}}). Therefore UDP throttling
+should be rather realised by per-flow policing, e.g, based on the hash
+of the UDP datagram's source and destination
+addresses and blocking a portion of the hash space that corresponds to the
+fraction of UDP traffic one wishes to drop.
+While QUIC endpoints are often able to survive address changes, e.g. by NAT rebindings,
+blocking a portion of the traffic based on 5-tuple hashing increases the risk of
+blackholing an active connection when the address changes.
```
mirjak, line 935:
```suggestion
While QUIC endpoints are often able to survive address changes, e.g. by NAT
rebindings, blocking a portion of the traffic based on 5-tuple hashing increases
the risk of black-holing an active connection when the address changes.
```

draft-ietf-quic-manageability.md
chris-box: Glad we are of the same mind (that a UDP cap is not ideal, but sometimes it may be necessary and if so here's how to do it).
My only comment on this sentence is to ask whether "adopted" should be "adapted". I would say that adoption means the process of starting to apply something, whereas adapted means changing it as you go. It's still possible to say you can adopt something in a dynamic fashion, but "adapted dynamically" is probably clearer.

draft-ietf-quic-manageability.md
chris-box: Looks good.

draft-ietf-quic-manageability.md
britram: I think you're both right. We need an example of how to do efficient per-flow policing here, but it could be Even More Clear that it's an example.

draft-ietf-quic-manageability.md
```diff
@@ -905,26 +905,36 @@ Admitting a few packets allows the QUIC endpoint to determine that the path
...
+blocked, QUIC-capable applications may failover to TCP
+However, blocking a
 random fraction of QUIC packets across 4-tuples will allow many QUIC handshakes
 to complete, preventing a TCP failover, but the connections will suffer from
-severe packet loss.
+severe packet loss (see also {{sec-filtering}}). Therefore UDP throttling
+should be rather realised by per-flow policing, e.g, based on the hash
+of the UDP datagram's source and destination
+addresses and blocking a portion of the hash space that corresponds to the
+fraction of UDP traffic one wishes to drop.
```
britram, line 933:
```suggestion
severe packet loss (see also {{sec-filtering}}). Therefore UDP throttling
should be realized by per-flow policing as opposed to per-packet
policing. Note that this per-flow policing should be stateless to avoid
problems with stateful treatment of QUIC flows (see {{sec-stateful}}),
for example blocking a portion of the space of values of a hash function
over the addresses and ports in the UDP datagram.
```

britram (approved):

draft-ietf-quic-manageability.md
```diff
@@ -905,26 +905,36 @@ Admitting a few packets allows the QUIC endpoint to determine that the path
...
+With increased deployment of QUIC, there is also an increased need to allow
+UDP traffic on ports used for QUIC. However, if UDP is generally enabled on
+these ports, UDP flood attacks may also use the same ports. One possible
+response to this threat is to throttle UDP traffic on the network, allocating a
+fixed portion of the network capacity to UDP and blocking UDP datagrams over
+that cap. As the portion of QUIC traffic compared to TCP is also expected to
+increase over time, using such a limit is not recommended but if done,
+limits might need to be adapted dynamically.
+
+Further, if UDP traffic is desired to be throttled, it is recommended to block individual
```
britram, line 923:
```suggestion
Further, if UDP traffic is desired to be throttled, it is recommended to
block individual
```

PR #306, opened 2021-03-25
Make recommendation to not CID for NAT stronger

### LINKED ISSUE #293: Stronger caution on using CID
@ianswett said in PR #272 on section "Passive Network Performance Measurement and Troubleshooting":

 "I think the caution against using Connection ID needs to be a lot stronger. Things will break, users will notice. I realize that implementations have to use this to enforce that, but I believe some already are. See a comment below as well."

mirjak:
fixes #293

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -840,7 +840,7 @@ communicate with the client.
 The QUIC header optionally contains a connection ID which can be used as
 additional entropy beyond the 5-tuple, if needed. The QUIC handshake needs
 to be observed in order to understand whether the connection ID is present and
-what length it has. However, connection IDs may be renegotiated
+what length it has. However, connection IDs may be renegotiated during
 after the handshake, and this renegotiation is not visible to the path. Using
 the connection ID as a flow key field for stateful treatment of flows may
 therefore cause undetectable and unrecoverable loss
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -840,7 +840,7 @@ communicate with the client.
 The QUIC header optionally contains a connection ID which can be used as
 additional entropy beyond the 5-tuple, if needed. The QUIC handshake needs
 to be observed in order to understand whether the connection ID is present and
-what length it has. However, connection IDs may be renegotiated during
+what length it has. However, connection IDs may be renegotiated
```
mirjak, line 843:
```suggestion
what length it has. However, connection IDs may be renegotiated during
```

draft-ietf-quic-manageability.md
```diff
@@ -849,9 +849,10 @@ discouraged for NAT applications.
 
 If a NAT hits an operational limit, it is recommended to rather drop the
```
ianswett, line 850:
```suggestion
If a NAT hits an operational limit, it is recommended to drop the
```

ianswett (approved):

draft-ietf-quic-manageability.md
mirjak: This is addressed in another PR.

draft-ietf-quic-manageability.md
```diff
@@ -849,9 +849,10 @@ discouraged for NAT applications.
...
 initial packets of a flow (see also {{sec-filtering}}),
-which potentially triggers a fallback to TCP, than using the connection ID to
-muliplex mulitple connections on the same IP address port pair which risks
-connectivity breakage later, in case the connection ID changes.
+which potentially triggers a fallback to TCP. Use of the connection ID to
+muliplex mulitple connections on the same IP address/port pair is not a
```
britram, line 853:
```suggestion
multiplex multiple connections on the same IP address/port pair is not a
```

britram (approved):

PR #308, opened 2021-03-26
Reference cleanup

MikeBishop:
Some section refs weren't linked to the sections; some objects were defined multiple ways; one reference wasn't actually cited in the text.


mirjak:
Thanks a lot!

PR #309, opened 2021-03-28
Remove details on using CID to associate flows

ianswett:
Currently I find this section worded in a way which could cause the reader to believe using CID would typically work.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -683,18 +683,17 @@ endpoints changes address or port.  This change can be due to NAT rebinding or
 address migration.
 
 The connection ID must change upon intentional address change by an endpoint,
-so it is only possible for a passive observer to link unintended changes of
-address using the connection ID.
+and connection ID negotiation is encrypted, so it is only possible for a
+passive observer to link unintended changes of address using the connection ID.
 
-In the case of the unintentional address change on a single endpoint
-(as is the case with NAT rebinding), an on-path observer can in most cases
+In the case of the unintentional address change on a single endpoint,
+as is the case with NAT rebinding, an on-path observer can in most cases
 associate the flow on the new address with the flow on the old address, using
 the connection ID.
 
-A network function that uses the connection ID to associate flows must
-be robust to the failure of this technique. Since the connection ID may
-change multiple times during the lifetime of a connection, and 
-connection ID negotiation is encrypted, packets with the
+A network function that attempts to use the connection ID to associate flows
+must be robust to the failure of this technique. Since the connection ID may
+change multiple times during the lifetime of a connection, packets with the
 same five-tuple but different connection IDs might or might not belong to
 the same connection. Likewise, packets with the same connection ID but
 different five-tuples might not belong to the same connection, either.
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -677,7 +677,7 @@ secret is retained by the observer.
 ## Flow Association {#sec-flow-association}
 
 The QUIC connection ID (see {{rebinding}}) is designed to allow a coordinating
-on-path device, such as a load-balancer, associate two flows when one of the
+on-path device, such as a load-balancer, to associate two flows when one of the
 endpoints changes address or port.  This change can be due to NAT rebinding or
 address migration.
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -676,18 +676,14 @@ secret is retained by the observer.
 
 ## Flow Association {#sec-flow-association}
 
-The QUIC connection ID (see {{rebinding}}) is designed to allow an on-path
-device such as a load-balancer to associate two flows as identified by
-five-tuple when the address and port of one of the endpoints changes; e.g. due
-to NAT rebinding or address migration. An observer keeping flow state
-can associate a connection ID, if present, with a given flow, and can associate
-a known flow with a new flow when when observing a packet sharing the same
-connection ID in the same direction between client and server and sharing one
-endpoint address (IP address and port) with the known flow.
+The QUIC connection ID (see {{rebinding}}) is designed to allow a coordinating
+on-path device, such as a load-balancer, associate two flows when one of the
+endpoints change address or port.  This change can be due to NAT rebinding or
+address migration.
 
-However, since the connection ID may change multiple times during the lifetime
-of a flow, and the negotiation of connection ID changes is encrypted, packets
-with the same 5-tuple but different connection IDs might or might not belong to
+Since the connection ID may change multiple times during the lifetime
+of a connection, and connection ID negotiation is encrypted, packets with the
+same five-tuple but different connection IDs might or might not belong to
 the same connection.
 
 Connection IDs should be treated as opaque; see {{sec-loadbalancing}}
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -681,6 +681,10 @@ on-path device, such as a load-balancer, associate two flows when one of the
 endpoints change address or port.  This change can be due to NAT rebinding or
 address migration.
 
+The connection ID must change upon intentional address change by an endpoint,
+so it is only possible for a passive observer to link unintended changes of
+address using the connection ID.
+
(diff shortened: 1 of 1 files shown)


mirjak:
While we say in multiple places that it is not recommended to to use the connection ID, it is still true that you can use it to associate a new 5-tuple to an existing connection. This is what this text describes and that's inline the with the goal of section 2 which basically only describes what possible with the wire image and what's not. Therefore I wouldn't want to remove this text here.

Note that there can be stateful functions like spin-bit based monitoring that would benefit from map connections with the same connection ID but don't break anything otherwise. 

However, I guess we could add a warning already here that it is usually not recommended to use the connection ID as it might change. I think that's what the next part already says but not directly as a warning or recommendation (given that was not the goal of this section).

martinthomson (approved):
I think that it is good to remove that text.  It is true that in some narrow cases the same connection ID on two paths means that the connection moved from one path to the other.  However, this is not a guarantee and to suggest otherwise is not very helpful.

I support removing the text.

ianswett:
I added another paragraph to clarify what is possible, PTAL

draft-ietf-quic-manageability.md
```diff
@@ -676,18 +676,18 @@ secret is retained by the observer.
...
-a known flow with a new flow when when observing a packet sharing the same
-connection ID in the same direction between client and server and sharing one
-endpoint address (IP address and port) with the known flow.
-
-However, since the connection ID may change multiple times during the lifetime
-of a flow, and the negotiation of connection ID changes is encrypted, packets
-with the same 5-tuple but different connection IDs might or might not belong to
+The QUIC connection ID (see {{rebinding}}) is designed to allow a coordinating
+on-path device, such as a load-balancer, associate two flows when one of the
+endpoints change address or port.  This change can be due to NAT rebinding or
```
britram, line 681:
```suggestion
endpoints changes address or port.  This change can be due to NAT rebinding or
```

draft-ietf-quic-manageability.md
```diff
@@ -676,18 +676,18 @@ secret is retained by the observer.
...
-can associate a connection ID, if present, with a given flow, and can associate
-a known flow with a new flow when when observing a packet sharing the same
-connection ID in the same direction between client and server and sharing one
-endpoint address (IP address and port) with the known flow.
-
-However, since the connection ID may change multiple times during the lifetime
-of a flow, and the negotiation of connection ID changes is encrypted, packets
-with the same 5-tuple but different connection IDs might or might not belong to
+The QUIC connection ID (see {{rebinding}}) is designed to allow a coordinating
+on-path device, such as a load-balancer, associate two flows when one of the
```
britram, line 680:
```suggestion
on-path device, such as a load-balancer, to associate two flows when one of the
```

draft-ietf-quic-manageability.md
```diff
@@ -676,18 +676,18 @@ secret is retained by the observer.
...
-of a flow, and the negotiation of connection ID changes is encrypted, packets
-with the same 5-tuple but different connection IDs might or might not belong to
+The QUIC connection ID (see {{rebinding}}) is designed to allow a coordinating
+on-path device, such as a load-balancer, to associate two flows when one of the
+endpoints changes address or port.  This change can be due to NAT rebinding or
+address migration.
+
+The connection ID must change upon intentional address change by an endpoint,
+so it is only possible for a passive observer to link unintended changes of
+address using the connection ID.
```
britram, line 686:
```suggestion
The connection ID must change upon intentional address change by an endpoint,
so it is only possible for a passive observer to link unintended changes of
address using the connection ID.

In the case of the unintentional address change on a single endpoint (as is the case with NAT rebinding), an on-path observer can in most cases associate the flow on the new address with the flow on the old address, using the connection ID.
```

draft-ietf-quic-manageability.md
```diff
@@ -676,18 +676,18 @@ secret is retained by the observer.
...
+
+Since the connection ID may change multiple times during the lifetime
+of a connection, and connection ID negotiation is encrypted, packets with the
+same five-tuple but different connection IDs might or might not belong to
 the same connection.
```
britram, line 691:
```suggestion
A network function that uses the connection ID to associate flows must
be robust to the failure of this technique. Since the connection ID may
change multiple times during the lifetime of a connection, and 
connection ID negotiation is encrypted, packets with the
same five-tuple but different connection IDs might or might not belong to
the same connection. Likewise, packets with the same connection ID but
different five-tuples might not belong to the same connection, either.
```

britram: @ianswett editing by editors led to this change (covering unintentional address change and the fact that observation techniques are brittle); wdyt?

ianswett: I'd prefer this paragraph without the first sentence, since that's reintroducing text suggesting a middlebox might do this.  If we are going to keep it, I think it needs to be reworded slightly.

The last sentence is a nice addition, though.

ianswett (commented):
I'm going to accept both changes and then make some follow-up changes.

draft-ietf-quic-manageability.md
```diff
@@ -677,19 +677,26 @@ secret is retained by the observer.
...
+The QUIC connection ID (see {{rebinding}}) is designed to allow a coordinating
+on-path device, such as a load-balancer, to associate two flows when one of the
+endpoints changes address or port.  This change can be due to NAT rebinding or
+address migration.
+
+The connection ID must change upon intentional address change by an endpoint,
+and connection ID negotiation is encrypted, so it is only possible for a
+passive observer to link unintended changes of address using the connection ID.
+
+In the case of the unintentional address change on a single endpoint,
```
ianswett, line 689: I personally wouldn't add this paragraph, but including it and letting you decide.

ianswett (commented):
PTAL.

draft-ietf-quic-manageability.md
```diff
@@ -677,19 +677,26 @@ secret is retained by the observer.
...
-the same connection.
+The QUIC connection ID (see {{rebinding}}) is designed to allow a coordinating
+on-path device, such as a load-balancer, to associate two flows when one of the
+endpoints changes address or port.  This change can be due to NAT rebinding or
+address migration.
+
+The connection ID must change upon intentional address change by an endpoint,
+and connection ID negotiation is encrypted, so it is only possible for a
+passive observer to link unintended changes of address using the connection ID.
+
```
mirjak, line 688:
```suggestion
The connection ID must change upon intentional address change by an endpoint,
and connection ID negotiation is encrypted.
```
I'd prefer to keep the sentence below but remove this half sentence and make one paragraph out of it.

mirjak:
```suggestion
The connection ID must change upon intentional address change by an endpoint,
and connection ID negotiation is encrypted, packets with the same 5-tuple but
different connection IDs might or might not belong to  the same connection.
```
Actually I would prefer to keep the original sentence here...

draft-ietf-quic-manageability.md
mirjak, line 689:
```suggestion
Connection ID negotiation is encrypted and the connection ID must change upon
intentional address change by an endpoint. Still, in the case of the unintentional
address change on a single endpoint,
```
I prefer to remove the previous half sentence and make one paragraph out of these two sentences.

draft-ietf-quic-manageability.md
```diff
@@ -677,19 +677,26 @@ secret is retained by the observer.
...
+In the case of the unintentional address change on a single endpoint,
+as is the case with NAT rebinding, an on-path observer can in most cases
+associate the flow on the new address with the flow on the old address, using
+the connection ID.
```
ianswett, line 692:
```suggestion
and connection ID negotiation is encrypted, so it is not possible for a
passive observer to link intended changes of address using the connection ID.

When one endpoint unintentionally changes its address, as is the case with NAT
rebinding, an on-path observer may be able to use the connection ID to
associate the flow on the new address with the flow on the old address.
```

mirjak: That works! Will commit this suggestion and withdraw mine. Text is better and much clearer now! Thanks!

mirjak (approved):
This is ready to merge now!

PR #310, opened 2021-04-09
Conn ID and load balancers

### LINKED ISSUE #243: No need to repeat connection ID requirements
[This text](https://quicwg.org/ops-drafts/draft-ietf-quic-manageability.html#section-2.6-2) repeats requirements from the transport spec unnecessarily.

> Server-generated connection IDs should seek to obscure any encoding, of routing identities or any other information. Exposing the server mapping would allow linkage of multiple IP addresses to the same host if the server also supports migration. Furthermore, this opens an attack vector on specific servers or pools.

It might be easier to just strike this.

mirjak:
fixes #243

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -457,19 +457,16 @@ supplied via encrypted frames (see {{Section 5.1 of QUIC-TRANSPORT}}).
 Therefore, observing a new connection ID does not necessary indicate a new
 connection.
 
-Server-generated connection IDs should seek to obscure any encoded routing
-identities or any other information. Exposing the server mapping would allow
-linkage of multiple IP addresses to the same host if the server also supports
-migration. Furthermore, this opens an attack vector on specific servers or
-pools.
-
-The best way to obscure an encoding is to appear random to observers, which is
-most rigorously achieved with encryption. Even when encrypted, a scheme could
-embed the unencrypted length of the connection ID in the connection ID itself,
-instead of remembering it.
-
-{{?QUIC_LB=I-D.ietf-quic-load-balancers}} further specifies possible algorithms
-to generate connection IDs at load balancers.
+{{?QUIC_LB=I-D.ietf-quic-load-balancers}} specifies algorithms for
+encoding the server mapping in a connection ID in order to share this 
+information with selected on-path devices such as load balancers. Server
+mappings should only be exposed to selected entities. Uncontrollied exposure
+would allow linkage of multiple IP addresses to the same host if the server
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -457,19 +457,16 @@ supplied via encrypted frames (see {{Section 5.1 of QUIC-TRANSPORT}}).
...
-
-The best way to obscure an encoding is to appear random to observers, which is
-most rigorously achieved with encryption. Even when encrypted, a scheme could
-embed the unencrypted length of the connection ID in the connection ID itself,
-instead of remembering it.
-
-{{?QUIC_LB=I-D.ietf-quic-load-balancers}} further specifies possible algorithms
-to generate connection IDs at load balancers.
+{{?QUIC_LB=I-D.ietf-quic-load-balancers}} specifies algorithms for
+encoding the server mapping in a connection ID in order to share this
```
mirjak, line 461:
```suggestion
encoding the server mapping in a connection ID in order to share this
```

draft-ietf-quic-manageability.md
```diff
@@ -457,19 +457,16 @@ supplied via encrypted frames (see {{Section 5.1 of QUIC-TRANSPORT}}).
...
+information with selected on-path devices such as load balancers. Server
+mappings should only be exposed to selected entities. Uncontrollied exposure
+would allow linkage of multiple IP addresses to the same host if the server
+also supports migration which opens an attack vector on specific servers or
+pools. The best way to obscure an encoding is to appear random to any other
+observers, which is most rigorously achieved with encryption. Even when
+encrypted, a scheme could embed the unencrypted length of the connection ID
+in the connection ID itself, instead of remembering it.
```
chris-box, line 469: Suggested tweaks to address the advice to network operators who are not also one of the endpoints, which I think is the intended scope here.
```suggestion
mappings should only be exposed to selected entities. Uncontrolled exposure
would allow linkage of multiple IP addresses to the same host if the server
also supports migration which opens an attack vector on specific servers or
pools. The best way to obscure an encoding is to appear random to any other
observers, which is most rigorously achieved with encryption. As a result
any attempt to infer information from specific parts of a connection ID is
unlikely to be useful.
```

martinthomson (commented):
This is fine, but it presumes a lot about the outcome of things that aren't completely resolved in QUIC-LB.  I think that I would like to see more of this left to QUIC-LB to deal with, especially if we are planning to publish this before QUIC-LB (which seems likely).  That might mean being a lot more aggressive about cutting text here.

PR #312, opened 2021-04-12
Initial endpoint DDoS cooperation text

### LINKED ISSUE #240: Endpoint cooperation for DoS mitigation
[Section 4.4](https://quicwg.org/ops-drafts/draft-ietf-quic-manageability.html#section-4.4) talks at length about detecting and mitigating against (D)DoS attacks, but this is written with the assumption of an uncooperative endpoint (likely to be, but unstated, a server).  All except the Retry text where the design specifically forces interventions to be cooperative.

It is entirely reasonable for DoS classification to be done with full knowledge of valid connection IDs or at least valid connection ID lengths.  If you allow for cooperation, the text might then explain how this might allow established flows to be separated from spoofed flows and new connection attempts.

britram:
fixes #240. thanks @gorryfair for the text.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -954,8 +954,8 @@ compliance with ciphersuites); client and application fingerprinting for
 inventory; and to provide alerts for network intrusion detection and other
 next generation firewall functions.
 
-Current practices in detection and mitigation of Distributed Denial of Service
-(DDoS) attacks generally involve classification of incoming traffic (as
+Current practices in detection and mitigation of DDoS
+attacks generally involve classification of incoming traffic (as
 packets, flows, or some other aggregate) into "good" (productive) and "bad"
 (DDoS) traffic, and then differential treatment of this traffic to forward only
 good traffic. This operation is often done in a separate specialized mitigation
@@ -971,17 +971,18 @@ as in {{sec-garbage}} and stateful tracking of QUIC traffic as in
 Note that the use of a connection ID to support connection migration renders
 5-tuple based filtering insufficient and requires more state to be maintained by
 DDoS defense systems. For the common case of NAT rebinding, where the client's
-address changes without the client's intent or knowedge, DDoS defense systems
+address changes without the client's intent or knowledge, DDoS defense systems
 can detect a change in the client's endpoint address by linking flows based on
 the server's connection IDs. However, QUIC's linkability resistance ensures that
 a deliberate connection migration is accompanied by a change in the connection
 ID. In this case, the connection ID can not be used to distinguish valid, active
 traffic from new attack traffic.
 
-Alternatively, when an encrypted transport is used, it is also possible for
-endpoints to directly support these security functions.
-Endpoints can cooperate with an on-path device directly by e.g. sharing
-information about connection IDs.
+It is also possible for
+endpoints to directly support security functions such as DoS
+classification and mitigation.
+Endpoints can cooperate with an in-network device directly by e.g.
+sharing information about connection IDs.
 
 Another potential method could use an
 on-path network device that relies on pattern inferences in the traffic and
@@ -989,7 +990,10 @@ heuristics or machine learning instead of processing observed header
 information.
 
 However, it is questionable whether connection migrations must be supported
-during a DDoS attack. If the connection migration is not visible to the network
+during a DDoS attack. However, for unintentional mitigations, as is usually the
+case for NAT rebindings, the connection ID can be used to track
+connections (see {{sec-flow-association}}). If the connection
+migration is not visible to the network
 that performs the DDoS detection, an active, migrated QUIC connection may be
 blocked by such a system under attack. As soon as the connection blocking is
 detected by the client, the client may rely on the fast resumption mechanism
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -945,6 +945,15 @@ severe packet loss.
 
 ## DDoS Detection and Mitigation {#sec-ddos-dec}
 
+On-path observation of the transport headers of packets can be used for various
+security functions. For example, Denial of Service (DOS) and Distributed DOS
+(DDOS) attacks against the infrastructure or against an endpoint can be
+detected and mitigated by characterising anomalous traffic.
+Other uses include support for security audits (e.g., verifying the
+compliance with ciphersuites); client and application fingerprinting for
+inventory; and to provide alerts for network intrusion detection and other
+next generation firewall functions.
+
 Current practices in detection and mitigation of Distributed Denial of Service
 (DDoS) attacks generally involve classification of incoming traffic (as
 packets, flows, or some other aggregate) into "good" (productive) and "bad"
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -779,7 +779,7 @@ not the network RTT.
 Since the spin bit logic at each endpoint considers only samples from packets
 that advance the largest packet number, signal generation itself is
 resistant to reordering. However, reordering can cause problems at an observer
-by causing spurious edge detection and therefore inaccurate (i.e., lower) RTT
+by causing spurious edge \ and therefore inaccurate (i.e., lower) RTT
```
mirjak, line 782:
```suggestion
by causing spurious edge detection and therefore inaccurate (i.e., lower) RTT
```

draft-ietf-quic-manageability.md
```diff
@@ -945,37 +945,59 @@ severe packet loss.
...
-attack. If the connection migration is not visible to the network that performs
-the DDoS detection, an active, migrated QUIC connection may be blocked by such a
-system under attack. As soon as the connection blocking is detected by the
-client, the client may rely on the fast resumption mechanism provided by
-QUIC. When clients migrate to a new path, they should be prepared for the
-migration to fail and attempt to reconnect quickly.
-
-TCP syncookies {{?RFC4937}} are a well-established method of mitigating some
+ID. In this case, the connection ID can not be used to distinguish valid, active
+traffic from new attacl traffic.
```
nibanks, line 979:
```suggestion
traffic from new attack traffic.
```

draft-ietf-quic-manageability.md
```diff
@@ -945,37 +945,59 @@ severe packet loss.
...
+
+Alternatively, when an encrypted transport is used, it is also possible for
+endpoints to directly support these security functions.
```
nibanks, line 982: I don't quite understand this sentence. For one, why "when an encrypted transport is used"? This document is purely about an encrypted transport, right? And then "these security functions" is also unclear to me.

draft-ietf-quic-manageability.md
```diff
@@ -945,37 +945,59 @@ severe packet loss.
...
+Endpoints can cooperate with an on-path device directly by e.g. sharing
+information about connection IDs.
+
+Another potential method could use an
+on-path network device that relies on pattern inferences in the traffic and
+heuristics or machine learning instead of processing observed header
+information.
+
+However, it is questionable whether connection migrations must be supported
```
nibanks, line 991: Is this referring to just explicit connection migration or do you include NAT rebinding as well? Either way, I think it should be explicitly called out. I think, ideally, NAT rebinding should continue to work, but I tend to agree full, explicit connection migration likely isn't completely necessary while under attack.

draft-ietf-quic-manageability.md
gorryfair: On the first: I think this is true for any encrypted transport, but yes it does apply to QUIC.  
On the functions: I thought this was: DOS;DDOS;Malware detection; NIDS; etc

draft-ietf-quic-manageability.md
```diff
@@ -945,37 +945,59 @@ severe packet loss.
...
+security functions. For example, Denial of Service (DOS) and Distributed DOS
+(DDOS) attacks against the infrastructure or against an endpoint can be
+detected and mitigated by characterising anomalous traffic.
+Other uses include support for security audits (e.g., verifying the
+compliance with ciphersuites); client and application fingerprinting for
+inventory; and to provide alerts for network intrusion detection and other
+next generation firewall functions.
+
 Current practices in detection and mitigation of Distributed Denial of Service
 (DDoS) attacks generally involve classification of incoming traffic (as
```
britram, line 958:
```suggestion
Current practices in detection and mitigation of DDoS
attacks generally involve classification of incoming traffic (as
```

draft-ietf-quic-manageability.md
```diff
@@ -945,37 +945,59 @@ severe packet loss.
...
+Efficient classification of this DDoS traffic in the mitigation environment
+is key to the success of this approach. Limited first-packet garbage detection
+as in {{sec-garbage}} and stateful tracking of QUIC traffic as in
+{{sec-stateful}} above may be useful during classification.
 
 Note that the use of a connection ID to support connection migration renders
 5-tuple based filtering insufficient and requires more state to be maintained by
-DDoS defense systems. For the common case of NAT rebinding, DDoS defense systems
+DDoS defense systems. For the common case of NAT rebinding, where the client's
+address changes without the client's intent or knowedge, DDoS defense systems
```
britram, line 974:
```suggestion
address changes without the client's intent or knowledge, DDoS defense systems
```

draft-ietf-quic-manageability.md
```diff
@@ -945,37 +945,59 @@ severe packet loss.
...
-system under attack. As soon as the connection blocking is detected by the
-client, the client may rely on the fast resumption mechanism provided by
-QUIC. When clients migrate to a new path, they should be prepared for the
-migration to fail and attempt to reconnect quickly.
-
-TCP syncookies {{?RFC4937}} are a well-established method of mitigating some
+ID. In this case, the connection ID can not be used to distinguish valid, active
+traffic from new attack traffic.
+
+Alternatively, when an encrypted transport is used, it is also possible for
```
britram, line 981:
```suggestion
It is also possible for
```

draft-ietf-quic-manageability.md
```diff
@@ -945,37 +945,59 @@ severe packet loss.
...
+endpoints to directly support these security functions.
```
britram, line 982:
```suggestion
endpoints to directly support security functions.
```

draft-ietf-quic-manageability.md
```diff
@@ -945,37 +945,59 @@ severe packet loss.
...
+Endpoints can cooperate with an on-path device directly by e.g. sharing
+information about connection IDs.
```
britram, line 984:
```suggestion
Endpoints can also cooperate with an on-path device directly by e.g.
sharing information about connection IDs.
```

draft-ietf-quic-manageability.md
```diff
@@ -945,37 +945,59 @@ severe packet loss.
...
+
+Another potential method could use an
+on-path network device that relies on pattern inferences in the traffic and
+heuristics or machine learning instead of processing observed header
+information.
+
+However, it is questionable whether connection migrations must be supported
+during a DDoS attack. If the connection migration is not visible to the network
```
mirjak, line 992:
```suggestion
during a DDoS attack. However, for unintentional mitigations, as usually the case
for NAT rebindings, the connection ID can easily be used to track connections.
If the connection migration is not visible to the network
```

draft-ietf-quic-manageability.md
```diff
@@ -945,37 +945,63 @@ severe packet loss.
...
+information.
+
+However, it is questionable whether connection migrations must be supported
+during a DDoS attack. However, for unintentional mitigations, as is usually the
+case for NAT rebindings, the connection ID can be used to track
+connections (see {{sec-flow-association}}). If the connection
+migration is not visible to the network
+that performs the DDoS detection, an active, migrated QUIC connection may be
+blocked by such a system under attack. As soon as the connection blocking is
+detected by the client, the client may rely on the fast resumption mechanism
```
martinthomson, line 999: Resumption might be fast, but still costs a lot more to handle than keeping a connection alive.  This won't produce a good outcome in the case of DoS.  I don't think that this is good advice at all.

martinthomson (commented):
I don't like where this was taken.  As @mirjak said, it is not responsive to my comments about cooperation.  It still firmly assumes that DoS mitigation is being performed without cooperation from endpoints.  That might be unavoidable for good reasons, but we should be clearer about the consequences of that AND we should recommend that DoS mitigation be performed with cooperation from endpoints.

PR #314, opened 2021-04-14
More explicit stream ordering guidance.

### LINKED ISSUE #286: Make it more clear that streams are not ordered
Based on the discussion in webtrans at IETF-110, I guess we could be more explicit that QUIC streams do not guarantee to be received in any particular order.

britram:
Fixes #286.

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -319,7 +319,7 @@ designing an application's use of streams:
 
 - A single stream provides ordering. If the application requires certain data to
 be received in order, that data should be sent on the same stream. There is
-no guarantee of transmission ordering across streams.
+no guarantee of reception ordering across streams.
 
 - Multiple streams provide concurrency. Data that can be processed
 independently, and therefore would suffer from head of line blocking if forced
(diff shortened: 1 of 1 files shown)


mirjak (approved):

draft-ietf-quic-applicability.md
```diff
@@ -318,7 +318,8 @@ HTTP/3 in {{QUIC-HTTP}}. There are a few general principles to apply when
 designing an application's use of streams:
 
 - A single stream provides ordering. If the application requires certain data to
-be received in order, that data should be sent on the same stream.
+be received in order, that data should be sent on the same stream. There is
+no guarantee of transmission ordering across streams.
```
LPardue, line 322: would it be more consistent to say that there is no guarantee of reception ordering across streams?

I think the transport caveat for applications using QUIC is that they might write N bytes to stream 0, and N bytes to stream 4 (in order) and expect that the recipient would receive them in that order.

Talking about transmission order is still a useful point but it is slightlly different. A QUIC implementation might allow ordered writes (stream 0 then stream 4, as above) but it might not provide any guarantee of transmission order. Perhaps you can add another sentence to mention transmission with a forward ref to the prioritization section?

britram: Yeah. Reception order is better.

```suggestion
be received in order, that data should be sent on the same stream. There is
no guarantee of reception ordering across streams.
```

martinthomson (approved):

PR #320, opened 2021-04-16
follow up on DDoS section changes

mirjak:
We recently added this sentence but the text two paragraphs up already says this. Here is my proposed change, however, maybe we can shorten this paragraph even further.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -1022,7 +1022,7 @@ However, it is questionable whether connection migrations must be supported
 during a DDoS attack. While unintended migration without a connection ID
 change can be more easily supported, it might be acceptable to not
 support migrations of active QUIC connections that are not visible to
-the network functions performing the DDoS detection. 
+the network functions performing the DDoS detection.
 As soon as the connection blocking is detected by the client,
 the client may be able to rely on the fast resumption mechanism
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -1018,13 +1019,12 @@ heuristics or machine learning instead of processing observed header
...
-migration is not visible to the network
-that performs the DDoS detection, an active, migrated QUIC connection may be
-blocked by such a system under attack. As soon as the connection blocking is
-detected by the client, the client may rely on the fast resumption mechanism
+during a DDoS attack. While unintended migration where the connection ID
+stays unchanged can be more easiliy supported, it might be acceptable that
+connection migrations of active QUIC connections that are not visible to
+the network that performs the DDoS detection may be blocked by such a system
+under attack. As soon as the connection blocking is detected by the client,
+the client may be able to rely on the fast resumption mechanism
```
britram, line 1027:
```suggestion
during a DDoS attack. While unintended migration without a connection ID
change can be more easily supported, it might be acceptable to not
support migrations of active QUIC connections that are not visible to
the network functions performing the DDoS detection. 
As soon as the connection blocking is detected by the client,
the client may be able to rely on the fast resumption mechanism
```

PR #321, opened 2021-04-19
Addition to filtering section to apply also for other than NAT

mirjak:


draft-ietf-quic-manageability.md
```diff
@@ -923,7 +923,8 @@ load-balancers and servers.
 
 ## Filtering Behavior {#sec-filtering}
 
-{{?RFC4787}} describes possible packet filtering behaviors that relate to NATs.
+{{?RFC4787}} describes possible packet filtering behaviors that relate to NATs
+but is often also used on other scenairos where packet filtering is desired.
```
britram, line 927:
```suggestion
{{?RFC4787}} describes possible packet filtering behaviors that relate to NATs
but is often also used is other scenarios where packet filtering is desired.
```

PR #322, opened 2021-04-21
Clean up the abstract a bit

britram:
Read this for the first time in a while and the document has drifted slightly.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -58,9 +58,10 @@ informative:
 --- abstract
 
 This document discusses manageability of the QUIC transport protocol, focusing
-on caveats impacting network operations involving QUIC traffic. Its intended
-audience is network operators, as well as content providers that rely on the use
-of QUIC-aware middleboxes, e.g. for load balancing.
+on the implications of QUIC's design and wire image on network operations 
+involving QUIC traffic. Its intended audience is network and service 
+operators who rely on the use of transport-aware network functions, e.g. for 
+security or load balancing, as well as the implementors of those functions.
 
 --- middle
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -59,9 +59,9 @@ informative:
 
 This document discusses manageability of the QUIC transport protocol, focusing
 on the implications of QUIC's design and wire image on network operations
-involving QUIC traffic. Its intended audience is network and service
-operators who rely on the use of transport-aware network functions and the
-implementors of those functions.
+involving QUIC traffic. Its intended audience is network operators and
+equipment vendors who rely on the use of transport-aware network
+functions and the implementors of those functions.
 
 --- middle
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -58,9 +58,9 @@ informative:
 --- abstract
 
 This document discusses manageability of the QUIC transport protocol, focusing
-on the implications of QUIC's design and wire image on network operations 
-involving QUIC traffic. Its intended audience is network and service 
-operators who rely on the use of transport-aware network functions, e.g. for 
+on the implications of QUIC's design and wire image on network operations
+involving QUIC traffic. Its intended audience is network and service
+operators who rely on the use of transport-aware network functions, e.g. for
 security or load balancing, as well as the implementors of those functions.
 
 --- middle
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -58,9 +58,10 @@ informative:
 --- abstract
 
 This document discusses manageability of the QUIC transport protocol, focusing
-on caveats impacting network operations involving QUIC traffic. Its intended
-audience is network operators, as well as content providers that rely on the use
-of QUIC-aware middleboxes, e.g. for load balancing.
+on the implications of QUIC's design and wire image on network operations
+involving QUIC traffic. Its intended audience is network and service
+operators who rely on the use of transport-aware network functions, e.g. for
+security or load balancing, as well as the implementors of those functions.
```
martinthomson, line 64:
```suggestion
operators who rely on the use of transport-aware network functions and the
implementors of those functions.
```

"e.g." in an abstract seems wrong to me.  I would say that using "for example" here is fraught because of the sheer diversity of transport-aware network functions.  Security, troubleshooting, capacity planning, load balancing, flow management, accounting, and more would be closer to a representative list, but I fear even that wouldn't be enough.

draft-ietf-quic-manageability.md
```diff
@@ -58,9 +58,10 @@ informative:
 --- abstract
 
 This document discusses manageability of the QUIC transport protocol, focusing
-on caveats impacting network operations involving QUIC traffic. Its intended
-audience is network operators, as well as content providers that rely on the use
-of QUIC-aware middleboxes, e.g. for load balancing.
+on the implications of QUIC's design and wire image on network operations
+involving QUIC traffic. Its intended audience is network and service
+operators who rely on the use of transport-aware network functions, e.g. for
```
mirjak, line 63: Not sure it is clear what is meant by "service operators"

draft-ietf-quic-manageability.md
mirjak, line 64: I don't think the addition of "as well as the implementors of those functions" is needed...

britram: Agreed, reluctantly.

britram: I wanted to call out specifically that vendors in the operations space (not just operators themselves) should pay attention to this. However, if that's implied, we can cut it.

draft-ietf-quic-manageability.md
mirjak: so maybe "network operators and network equipment vendors"...? However, not sure we really need to call this out separately...

draft-ietf-quic-manageability.md
mirjak: I think we can cut this but I would have preferred to keep the examples.I don't see any problems with having e.g. in the abstract.

draft-ietf-quic-manageability.md
```diff
@@ -58,9 +58,10 @@ informative:
 --- abstract
 
 This document discusses manageability of the QUIC transport protocol, focusing
-on caveats impacting network operations involving QUIC traffic. Its intended
-audience is network operators, as well as content providers that rely on the use
-of QUIC-aware middleboxes, e.g. for load balancing.
+on the implications of QUIC's design and wire image on network operations
+involving QUIC traffic. Its intended audience is network and service
+operators who rely on the use of transport-aware network functions and the
+implementors of those functions.
```
britram, line 64:
```suggestion
on the implications of QUIC's design and wire image on network operations
involving QUIC traffic. Its intended audience is network operators and
equipment vendors who rely on the use of transport-aware network
functions and the implementors of those functions.
```

draft-ietf-quic-manageability.md
```diff
@@ -58,9 +58,10 @@ informative:
 --- abstract
 
 This document discusses manageability of the QUIC transport protocol, focusing
-on caveats impacting network operations involving QUIC traffic. Its intended
-audience is network operators, as well as content providers that rely on the use
-of QUIC-aware middleboxes, e.g. for load balancing.
+on the implications of QUIC's design and wire image on network operations
+involving QUIC traffic. Its intended audience is network operators and
+equipment vendors who rely on the use of transport-aware network
+functions and the implementors of those functions.
```
britram, line 64:
```suggestion
on the implications of QUIC's design and wire image on network operations
involving QUIC traffic. Its intended audience is network operators and
equipment vendors who rely on the use of transport-aware network
functions.
```

PR #336, opened 2021-05-06
same 5-tuple, same connection

martinthomson:
Probably, but not certain

Especially if you got it wrong regarding the connection ID.


mirjak:
If there is no connection is how would an endpoint identify connection on the same 5-tuple then (for QUICv1)?

martinthomson:
It is possible to run multiple connections on the same tuple with trial decryption.

ianswett (approved):

PR #347, opened 2021-05-06
Strike discussion of lower and higher timeouts

### LINKED ISSUE #346: Are we recommending shorter timeouts for middleboxes?
> [RFC4787] requires a timeout that is not less than 2 minutes for most UDP traffic. However, in practice, timers are sometimes lower, in the range of 30 to 60 seconds.

As this is advice for network operators and middlebox vendors, this reads to me as permission to use a lower timeout.  I think that it still makes sense to maintain the recommendation in RFC 4787.  The advice for applications (don't rely on 2 minutes) still applies, but a 2 minute recommendation is worth sticking to.

martinthomson:
Closes #346.


britram:
merging #350 instead

PR #373, opened 2021-05-19
Editorial edits based on issues

### LINKED ISSUE #329: Section 4.5 of applicability needs paragraph breaks
That mega paragraph has a bunch of good concepts, but it's all run together into a single blob, which makes it a bit hard to follow.

### LINKED ISSUE #367: Applicability: Trying to avoid confusion when read in future years
Three places where the text might avoid confusion when read in future years:

1) “Further downgrades to older TLS versions than used in QUIC, which is 1.3, “
… will it always be 1.3, or should this be “in QUIC [REF] is TLS 1.3”

2) “Currently, QUIC only provides fully reliable stream transmission, which means that prioritization of retransmissions will be beneficial in most cases, by filling in gaps and freeing up the flow control window. “
- I'm not so happy with currently here: Is this the best way to say this, might it be better to say this rather than even hinting that future QUIC might only be reliable:
“When a QUIC endpoint provides provides fully reliable stream transmission, prioritization of retransmissions will be beneficial in most cases, by filling in gaps and freeing up the flow control window. “

3)  I’d prefer to see a ref here, since this something that the WG might later decide to work upon:
“Currently QUIC only supports failover cases [Quic-Transport].”

### LINKED ISSUE #368: Applicability: Add a cross ref to the network part in manageability?
“QUIC exposes some information to the network in the unencrypted part of the header, either before the encryption context is established or because the information is intended to be used by the network.”
- Could this benefit form a cross-reference to manageability, where this topic is also discussed from a network perspective?

mirjak:
fixes #329, fixes #368, fixes #367

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -137,7 +137,8 @@ insecure protocols or to weaker versions of secure protocols.
 An application that implements fallback needs to consider the security
 consequences. A fallback to TCP and TLS exposes control information to
 modification and manipulation in the network. Further downgrades to older TLS
-versions than used in QUIC, which is 1.3, might result in significantly weaker
+versions than 1.3, which is used in QUIC version 1, might result in
+significantly weaker
 cryptographic protection. For example, the results of protocol negotiation
 {{?RFC7301}} only have confidentiality protection if TLS 1.3 is used.
 
@@ -379,10 +380,10 @@ application on top.
 
 Priority handling of retransmissions can be implemented by the sender in the
 transport layer. {{QUIC}} recommends retransmitting lost data before new data,
-unless indicated differently by the application. Currently, QUIC only provides
-fully reliable stream transmission, which means that prioritization of
-retransmissions will be beneficial in most cases, by filling in gaps and freeing
-up the flow control window. For partially reliable or unreliable streams,
+unless indicated differently by the application. When a QUIC endpoint uses
+fully reliable streams for transmission, prioritization of retransmissions will
+be beneficial in most cases, as filling in gaps and freeing up the flow
+control window. For partially reliable or unreliable streams,
 priority scheduling of retransmissions over data of higher-priority streams
 might not be desirable. For such streams, QUIC could either provide an
 explicit interface to control prioritization, or derive the prioritization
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-applicability.md
```diff
@@ -379,10 +380,10 @@ application on top.
 
 Priority handling of retransmissions can be implemented by the sender in the
 transport layer. {{QUIC}} recommends retransmitting lost data before new data,
-unless indicated differently by the application. Currently, QUIC only provides
-fully reliable stream transmission, which means that prioritization of
-retransmissions will be beneficial in most cases, by filling in gaps and freeing
-up the flow control window. For partially reliable or unreliable streams,
+unless indicated differently by the application. When a QUIC endpoint uses
+fully reliable streams for transmission, prioritization of retransmissions will
+be beneficial in most cases, as filling in gaps and freeing up the flow
```
ianswett, line 385: I prefer either "by" as you had it before or drop "as" entirely.

draft-ietf-quic-applicability.md
```diff
@@ -476,24 +477,26 @@ be opened, which prevents applications using QUIC from making further progress.
...
 to stop the server to perform scheduled maintenance. Immediate connection close
 causes abrupt closure of actively used streams. Depending on how an application
 uses QUIC streams, this could be undesirable or detrimental to behavior or
-performance. A more graceful closure technique is to stop sending increases to
+performance.
+
+A more graceful closure technique is to stop sending increases to
 stream limits and allow the connection to naturally terminate once remaining
 streams are consumed. However, the period of time it takes to do so is dependent
 on the client and an unpredictable closing period might not fit application or
```
ianswett, line 490: I don't think this is client-only in general, even if it is for HTTP/3.

```suggestion
on the peer and an unpredictable closing period might not fit application or
```

draft-ietf-quic-applicability.md
```diff
@@ -683,7 +687,9 @@ application-layer information provided.
 
 QUIC exposes some information to the network in the unencrypted part of the
 header, either before the encryption context is established or because the
-information is intended to be used by the network. QUIC has a long header that
+information is intended to be used by the network. For more information on
+manageability of QUIC ofsee also {{?I-D.ietf-quic-manageability}}.
```
ianswett, line 691:
```suggestion
manageability of QUIC see also {{?I-D.ietf-quic-manageability}}.
```

ianswett (approved):

draft-ietf-quic-applicability.md
mirjak:
```suggestion
be beneficial in most cases, filling in gaps and freeing up the flow
```

gorryfair:
From PR #378: I'd also like to see something like :
/In this section, we discuss those aspects of the QUIC transport protocol that/This section discusses those aspects of the QUIC transport protocol that/
/Here, we are concerned primarily with the unencrypted part of QUIC's wire image
 {{WIRE-IMAGE}}, which we define as the information available in the packet/This section is concerned primarily with the unencrypted part of QUIC's wire image
 {{WIRE-IMAGE}}, defined as the information available in the packet/

and
/least four datagrams we'll call "Client Initial",/least four datagrams, labelled "Client Initial",/

and
/ 0-RTT data may also be
 seen in the Client Initial datagram, as shown in /When the client uses 0-RTT connection resumption, the Client Initial datagram can also carry 0-RTT data, as shown in/
and
/typically around 10 packets/typically to  around 10 packets/
and
/of a connection after one of one endpoint/of a connection after an endpoint/

I'd prefer more like:
/Though the guidance there holds, a particularly unwise behavior admits a
 handful of UDP packets and then makes a decision as to whether or not to filter
 later packets in the same connection. 
 QUIC applications are encouraged to fail over to TCP if early packets do/
... and I'll close PR #378.

Issue #364, opened 2021-05-12
Manageability: does the endpoint change address using a NAPT?


gorryfair:
In this text:
“When one endpoint unintentionally changes its address, as is the case with NAT rebinding, “
- actually the endpoint doesn’t change the address, except in RSIP (odd to find that in use!), it is the “received endpoint address” that changes,.

britram:
passivize, perhaps: "when one endpoint's address unintentionally changes?"

Issue #362, opened 2021-05-12
Manageability: Detailed Review Comments - NiTs


gorryfair:
This issue captires a number of minor wording issues, references etc:

Operators should expect to observe packets with other version numbers as a result of various Internet experiments, future standards, and greasing.
- Since greasing might not be understood by all operations staff, a reference here would help, e.g.: RFC8701.

Style:
“four datagrams we'll call “
- why suddenly a use of “we”, can you reword to remove that?

This could be clearer:
“When the client uses 0-RTT connection resumption, 0-RTT data may also be seen in the Client Initial datagram, as shown in Figure 6.”
- what does “seen” mean? is “can be present” equivalent, in which case it is better than implying it is observable.

“will result in connection establishment failing later on.”
- Can we just remove the spurious word “on” at the end of the sentence?

“This allows rebinding of a connection after one of one endpoint experienced an address change”
- what is “one of one”?

“also supports migration which opens an attack vector on specific servers or pools.”
- add comma before “which” or use “that”?

Section 2.7 could refer to https://datatracker.ietf.org/doc/draft-ietf-tsvwg-transport-encrypt, (now in RFC-Ed) for examples of sequence number observation and alternatives for encrypted transports.

Section 3:
“This section addresses the different kinds of observations and inferences that can be made about QUIC flows by a passive observer in the network based on the wire image in Section 2.”
- This seems correct, although may be helpful to also say this observer does not have access to keying information.

Actually the set of possible things that can be done is more than the say, please add a reference
for example by queueing ACKs differently or manipulating ACK....
- It might be worth citing RFC3449 as example? or https://datatracker.ietf.org/doc/draft-ietf-tsvwg-transport-encrypt.
- Although other things also are done and also some links compress ACKs, etc there is a more modern view in:  per https://datatracker.ietf.org/doc/draft-ietf-tsvwg-transport-encrypt./.

The title here: 4.8. Quality of Service handling and ECMP 
- should this title be “ 4.8. Quality of Service handling and ECMP Routing”
- just to complete the concept of ECMP in the section title?

gorryfair:
And ... 
The text says: “a particularly unwise behavior is to admit a handful of UDP packets and then make a decision as to whether or not to filter it”
- “it” is presumably “later packets in the flow”?

britram:
...check first to make sure we didn't already merge a PR for these...

mirjak:
[...]

Issue #342, opened 2021-05-06
Is identification of version 1 ossification?


martinthomson:
Is this a promise?

> An in-network observer assuming that a set of packets belongs to a QUIC flow can infer the version number in use by observing the handshake: for QUIC version 1, if the version number in the Initial packet from a client is the same as the version number in the Initial packet of the server response, that version has been accepted by both endpoints to be used for the rest of the connection.

Because there are a number of designs floating around that would allow a client and server to agree - out of band or by prior arrangement - that version 1 means something else entirely.

If the above were a promise, that's perhaps reasonable.  It is an interpretation that seems consistent with the invariants.  However, the invariants don't make this particular (that was deliberate), so this is technically stretching it.

This might be better phrased with a "might infer the version number in use" to admit the possibility of error.

PR #374, opened 2021-05-26
Nits and smaller editorial changes based on issues

### LINKED ISSUE #354: Are client-requested Connection IDs atypical?
> Client and server negotiate connection IDs during the handshake; typically, however, only the server will request a connection ID for the lifetime of the connection.

From a discussion on this assertion with @nibanks, he observed that while client-side CIDs are uncommon on "actual" clients, they're likely to be very common on proxies (e.g. Masque).  Anyone monitoring traffic at the server network might well see lots of client-side CIDs.

That suggests this statement about what "typically" happens might be overly specific to one deployment or viewpoint.

mirjak:
fixes #354 #342 #362 #364

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -1169,8 +1169,8 @@ network as well as attacks on specific QUIC mechanism.
 Version Negotiation packets do not contain any mechanism to prevent version
 downgrade attacks. However, future versions of QUIC that use Version Negotiation
 packets are required to define a mechanism that is robust against version
-downgrade attacks. Therefore, a network node should not attempt to impact version
-selection, as version downgrade may result in connection failure.
+downgrade attacks. Therefore, a network node should not attempt to impact
+version selection, as version downgrade may result in connection failure.
 
 # Contributors
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -299,7 +299,7 @@ Initial packet, or leaving unused payload in the UDP packet after the Initial
 packet. A network path needs to be able to forward at least this size of
 packet for QUIC to be used.
 
-The content of Initial packets are encrypted using Initial Secrets, which
+The content of Initial packets is encrypted using Initial Secrets, which
 are derived from a per-version constant and the client's destination connection
 ID; they are therefore observable by any on-path device that knows the
 per-version constant. They are therefore considered visible in this
@@ -368,7 +368,7 @@ obfuscated using the Initial secret.
 +------------------------------------------------------------+
 | QUIC long header (type = Initial, Version, DCID, SCID)   (Length)
 +------------------------------------------------------------+  |
-| QUIC ACK frame (acknowledging Server Initial Initial)        |  |
+| QUIC ACK frame (acknowledging Server Initial)              |  |
 +------------------------------------------------------------+<-+
 | QUIC long header (type = Handshake, Version, DCID, SCID) (Length)
 +------------------------------------------------------------+  |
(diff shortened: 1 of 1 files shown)


mirjak:
#362 still has two open points ("is RFC8701 the right reference for greasing" and "should we refer draft-ietf-tsvwg-transport-encrypt"). However we might just open two new issue for those if needed...

mirjak:
Comments from @gorryfair (but were posted on the wrong PR):

From PR #378: I'd also like to see something like :
/In this section, we discuss those aspects of the QUIC transport protocol that/This section discusses those aspects of the QUIC transport protocol that/
/Here, we are concerned primarily with the unencrypted part of QUIC's wire image
{{WIRE-IMAGE}}, which we define as the information available in the packet/This section is concerned primarily with the unencrypted part of QUIC's wire image
{{WIRE-IMAGE}}, defined as the information available in the packet/

and
/least four datagrams we'll call "Client Initial",/least four datagrams, labelled "Client Initial",/

and
/ 0-RTT data may also be
seen in the Client Initial datagram, as shown in /When the client uses 0-RTT connection resumption, the Client Initial datagram can also carry 0-RTT data, as shown in/
and
/typically around 10 packets/typically to around 10 packets/
and
/of a connection after one of one endpoint/of a connection after an endpoint/

I'd prefer more like:
/Though the guidance there holds, a particularly unwise behavior admits a
handful of UDP packets and then makes a decision as to whether or not to filter
later packets in the same connection.
QUIC applications are encouraged to fail over to TCP if early packets do/
... and I'll close PR #378.

draft-ietf-quic-manageability.md
```diff
@@ -102,15 +102,15 @@ protocol as defined.
...
-version to version. However, the field that identifies the QUIC version in some
-packets, and the format of the Version Negotiation Packet, are both inspectable
-and invariant {{?QUIC-INVARIANTS=I-D.ietf-quic-invariants}}.
+This section is therefore primarily considering the unencrypted part of QUIC's
+wire image {{WIRE-IMAGE}}, which is defined as the information available in the
+packet header in each QUIC packet, and the dynamics of that information. Since
+QUIC is a versioned protocol, the wire image of the header format can also change
+from version to version. However, the field that identifies the QUIC version in
+some packets, and the format of the Version Negotiation Packet, are both
+inspectable and invariant {{?QUIC-INVARIANTS=I-D.ietf-quic-invariants}}.
```
mirjak, line 113:
```suggestion
QUIC is a versioned protocol, the wire image of the header format can also
change from version to version. However, the field that identifies the QUIC
version in some packets, and the format of the Version Negotiation Packet,
are both inspectable and invariant
{{?QUIC-INVARIANTS=I-D.ietf-quic-invariants}}.
```

mirjak:
Added Gorry's additional comments. This should be ready to go!

draft-ietf-quic-manageability.md
```diff
@@ -1165,8 +1168,8 @@ network as well as attacks on specific QUIC mechanism.
 
 Version Negotiation packets do not contain any mechanism to prevent version
 downgrade attacks. However, future versions of QUIC that use Version Negotiation
-packets are require to define a mechanism that is robust against version
-downgrade attacks. Therefore a network node should not attempt to impact version
+packets are required to define a mechanism that is robust against version
+downgrade attacks. Therefore, a network node should not attempt to impact version
 selection, as version downgrade may result in connection failure.
```
mirjak, line 1173:
```suggestion
downgrade attacks. Therefore, a network node should not attempt to impact
version selection, as version downgrade may result in connection failure.
```

PR #375, opened 2021-05-26
More explanation about the APLN token "h3"

### LINKED ISSUE #360: 1 ALPN per version.
Applicability Sec 8:

>Note that HTTP/3's ALPN token ("h3") identifies not only the version of the application protocol, but also the version of QUIC >itself; this approach allows unambiguous agreement between the endpoints on the protocol stack in use.

I was going to start a discussion about this at IETF 111. I think we should think hard about this; if there are a lot of versions with small differences to the application, this could result in a ridiculous cloud of RFCs.

mirjak:
fixes #360

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -595,7 +595,7 @@ Applications could define an alternate endpoint discovery mechanism to allow
 the usage of ports other than the default. For example, HTTP/3 ({{Sections 3.2
 and 3.3 of QUIC-HTTP}}) specifies the use of HTTP Alternative Services
 {{?RFC7838}} for an HTTP origin to advertise the availability of an equivalent
-HTTP/3 endpoint on a certain UDP port by using the "h3" ALPN token.  
+HTTP/3 endpoint on a certain UDP port by using the "h3" ALPN token.
 
 Since QUIC version 1 deferred defining a version negotiation mechanism, HTTP/3
 also deferred deciding how it would map to other QUIC versions. HTTP/3 requires
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -601,11 +601,12 @@ Since QUIC version 1 deferred defining a version negotiation mechanism, HTTP/3
 also deferred deciding how it would map to other QUIC versions. HTTP/3 requires
 QUIC version 1 and defines the ALPN token ("h3") to only apply to that version;
 this approach allows unambiguous agreement between the endpoints on the protocol
-stack in use but solves the versioning problem at the cost of future flexibility.
-Application protocol mappings written for QUIC version 1 or other QUIC versions
-may be able to benefit from version negotiation mechanisms that did not exist
-during the development of HTTP/3 . Coupling QUIC version to ALPN tokens is one
-approach but not the only approach.
+stack in use but solves the versioning problem at the cost of future
+flexibility. Application protocol mappings written for QUIC version 1 or
+other QUIC versions may be able to benefit from version negotiation
+mechanisms that did not exist during the development of HTTP/3.
+Coupling QUIC version to ALPN tokens is one approach but not the
+only approach.
 
 Given the prevalence of the assumption in network management
 practice that a port number maps unambiguously to an application, the
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -599,22 +599,18 @@ and 3.3 of QUIC-HTTP}}) specifies the use of HTTP Alternative Services
 HTTP/3 endpoint on a certain UDP port by using the "h3" Application-Layer
 Protocol Negotiation (ALPN) {{?RFC7301}} token.
 
-Additionally, ALPN permits the
+ALPN permits the
 client and server to negotiate which of several protocols will be used on a
 given connection.  Therefore, multiple applications might be supported on a
 single UDP port based on the ALPN token offered.  Applications using QUIC
-should register an ALPN token for use in the TLS handshake.
+are required to register an ALPN token for use in the TLS handshake.
 
 As QUIC version 1 deferred defining a complete version negotiation mechanism,
 HTTP/3 requires QUIC version 1 and defines the
-ALPN token ("h3") to only apply to that version. This approach avoids
-ambiguity of the used version in absence of a version negotiation
-mechanism. So far no single approach has been selected for
+ALPN token ("h3") to only apply to that version. 
+So far no single approach has been selected for
 managing the use of different QUIC versions, neither in HTTP/3 nor in general.
-Coupling QUIC version to ALPN tokens is one approach but not the
-only approach. Other options may be suitable when a version negotiation
-scheme gets developed and new protocol versions (at both application and
-transport layers) are defined. Application protocols that use QUIC need to
+Application protocols that use QUIC need to
 consider how the protocol will manage different QUIC versions.
 Decisions for those protocols might be informed by choices made by other
 protocols, like HTTP/3.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-applicability.md
```diff
@@ -595,10 +595,17 @@ Applications could define an alternate endpoint discovery mechanism to allow
 the usage of ports other than the default. For example, HTTP/3 ({{Sections 3.2
 and 3.3 of QUIC-HTTP}}) specifies the use of HTTP Alternative Services
 {{?RFC7838}} for an HTTP origin to advertise the availability of an equivalent
-HTTP/3 endpoint on a certain UDP port by using the "h3" ALPN token.  Note that
-HTTP/3's ALPN token ("h3") identifies not only the version of the application
-protocol, but also the version of QUIC itself; this approach allows unambiguous
-agreement between the endpoints on the protocol stack in use.
+HTTP/3 endpoint on a certain UDP port by using the "h3" ALPN token.
```
mirjak, line 598:
```suggestion
HTTP/3 endpoint on a certain UDP port by using the "h3" ALPN token.
```

martinthomson (commented):
We could say a lot or a little about this.  This is a lot.

Saying much less would be my preference.  The more we say, the worse it becomes when viewed from the future.

I would stick to the central facts:

* QUIC version 1 did not define a complete version negotiation system.
* HTTP/3 is defined as using only QUIC version 1.
* No single approach has been selected for managing the use of different QUIC versions, either in HTTP/3 or in general.
* This could change as the version negotiation scheme is developed and new protocol versions (at both application and transport layers) are defined.

The trick here is in deciding what to tell people who are building new application protocols.  The only answer I have there is "watch what HTTP does".  That isn't necessarily what you want to write down though, but if you think that you need advice, then maybe that's OK:

> Application protocols that use QUIC need to consider how the protocol will manage different QUIC versions.  Decisions for those protocols might be informed by choices made by other protocols, like HTTP/3.

draft-ietf-quic-applicability.md
```diff
@@ -585,28 +585,40 @@ TCP port already registered for the application is appropriate. For example,
...
+ambiguity of the used version in absence of a version negotiation
+mechanism. So far no single approach has been selected for
+managing the use of different QUIC versions, neither in HTTP/3 nor in general.
+Coupling QUIC version to ALPN tokens is one approach but not the
+only approach. Other options may be suitable when a version negotiation
+scheme gets developed and new protocol versions (at both application and
+transport layers) are defined. Application protocols that use QUIC need to
+consider how the protocol will manage different QUIC versions.
+Decisions for those protocols might be informed by choices made by other
+protocols, like HTTP/3.
```
martinduke, line 620: I propose that you simply delete this paragraph.

First, I disagree that there is "ambiguity of the used version." The version field is quite explicit; the only thing QUIC is missing is downgrade detection.

More broadly, I don't think we have a consensus on how ALPN will work with QUIC versions, so it is not possible to say anything useful here, unless we're going to have that conversation before this ships.

The rest of the new text is great.

martinduke (changes requested):

draft-ietf-quic-applicability.md
mirjak: This is all of the new text (I just move the other paragraphs around to make things fit). If we don't want this new part, I would like to keep this sentence from the old text instead because I think that's the important bit of information:

`Note that HTTP/3's ALPN token ("h3") identifies not only the version of the application protocol, but also the version of QUIC itself`

We can remove the part about ambiguity...

martinduke: Well this is the part that I disagree with. I would like QUICv2 to also use the h3 token, and don't see why that's incompatible with the text in base-drafts.

draft-ietf-quic-applicability.md
```diff
@@ -585,28 +585,40 @@ TCP port already registered for the application is appropriate. For example,
...
 identification.
 
+Applications could define an alternate endpoint discovery mechanism to allow
+the usage of ports other than the default. For example, HTTP/3 ({{Sections 3.2
+and 3.3 of QUIC-HTTP}}) specifies the use of HTTP Alternative Services
+{{?RFC7838}} for an HTTP origin to advertise the availability of an equivalent
+HTTP/3 endpoint on a certain UDP port by using the "h3" Application-Layer
+Protocol Negotiation (ALPN) {{?RFC7301}} token.
+
+Additionally, ALPN permits the
```
martinthomson, line 602:
```suggestion
ALPN permits the
```

draft-ietf-quic-applicability.md
```diff
@@ -585,28 +585,40 @@ TCP port already registered for the application is appropriate. For example,
...
+client and server to negotiate which of several protocols will be used on a
+given connection.  Therefore, multiple applications might be supported on a
+single UDP port based on the ALPN token offered.  Applications using QUIC
+should register an ALPN token for use in the TLS handshake.
```
martinthomson, line 606:
```suggestion
are required to register an ALPN token for use in the TLS handshake.
```

ALPN - or something like it - is mandatory in QUIC.  But when TLS is used - and this is about QUIC v1 where TLS is used - that effectively means ALPN is mandatory.

draft-ietf-quic-applicability.md
```diff
@@ -585,28 +585,40 @@ TCP port already registered for the application is appropriate. For example,
...
+
+As QUIC version 1 deferred defining a complete version negotiation mechanism,
+HTTP/3 requires QUIC version 1 and defines the
+ALPN token ("h3") to only apply to that version. This approach avoids
+ambiguity of the used version in absence of a version negotiation
+mechanism. So far no single approach has been selected for
```
martinthomson, line 612:
```suggestion
ALPN token ("h3") to only apply to that version. 
So far no single approach has been selected for
```
The ambiguity exists as long as there is even the potential for another QUIC version.  Safer not to say anything.

draft-ietf-quic-applicability.md
```diff
@@ -585,28 +585,40 @@ TCP port already registered for the application is appropriate. For example,
...
+managing the use of different QUIC versions, neither in HTTP/3 nor in general.
+Coupling QUIC version to ALPN tokens is one approach but not the
+only approach. Other options may be suitable when a version negotiation
+scheme gets developed and new protocol versions (at both application and
+transport layers) are defined. Application protocols that use QUIC need to
```
martinthomson, line 617:
```suggestion
Application protocols that use QUIC need to
```

I don't see there being any value in talking about this coupling further.

draft-ietf-quic-applicability.md
martinthomson: I think that there is value in the statement; that it is currently true is useful information, but the text doesn't say that it is necessarily going to stay that way.

If the objection is to the implication, would it be more acceptable if it said "HTTP/3 requires QUIC version 1 and defines the ALPN token ("h3") to only apply to that version.  This decision could be revised when new QUIC versions are defined."

martinthomson (commented):
I can make it shorter.

Maybe not as short as @martinduke would have it, but still shorter :)

draft-ietf-quic-applicability.md
martinduke: While I think we have a lot of basic decisions still to make about ALPN and versioning, the warning at the end of your proposed text does make it clear that this is all in flux. I can live with it.

draft-ietf-quic-applicability.md
```diff
@@ -585,28 +585,36 @@ TCP port already registered for the application is appropriate. For example,
...
+
+ALPN permits the
+client and server to negotiate which of several protocols will be used on a
+given connection.  Therefore, multiple applications might be supported on a
+single UDP port based on the ALPN token offered.  Applications using QUIC
+are required to register an ALPN token for use in the TLS handshake.
+
+As QUIC version 1 deferred defining a complete version negotiation mechanism,
+HTTP/3 requires QUIC version 1 and defines the
+ALPN token ("h3") to only apply to that version.
```
mirjak, line 610:
```suggestion
ALPN token ("h3") to only apply to that version.
```

Issue #344, opened 2021-05-06
Parsing Initials to get the DCID


martinthomson:
... doesn't make it clear that only the Initial from the client (and only those sent before receiving packets from the server) contain a Destination Connection ID field that can be used to decrypt the packet.  Any initial from the server, or any initial sent by a client after receiving Initial packets from the server, probably won't work.

martinthomson:
Hmm, this is the last paragraph.  Too late?

martinduke:
More broadly, it would be nice if Sec 2.4 discussed the property that the server CID changes pretty much immediately. We wouldn't want people to bind to a 2-tuple of CIDs and have that immediately fail.

mirjak:
@martinduke I create PR #376 based on your comment. Is that sufficient?

@martinthomson which section are you talking about?

britram:
I think #380 will fix @martinthomson's comment here.

PR #376, opened 2021-05-26
note that DCID change after client initial

mirjak:
this addresses a comment from @martinduke in issue #344


britram:
In the interests of timely completion of the document, will go ahead and merge this; please open an issue if this doesn't work for you @martinduke :)

PR #379, opened 2021-05-26
More aspects of setting DSCPs

gorryfair:


draft-ietf-quic-manageability.md
```diff
@@ -1055,18 +1055,28 @@ It is expected that any QoS handling in the network, e.g. based on use of
...
 
+If a packet enters a network segment that does not support the DSCP value, 
+this could result in the connection not receiving the network treatment 
+it expects. The DSCP value in this packet could also be remarked as the 
+packet travels along the network path, changing the requested treatment.
+
+When multiplexing multiple flows over a QUIC connection, the selected 
+DSCP value should be the one associated with the highest priority requested 
+for all multiplexed flows.
+
```
mirjak, line 1079:
```suggestion
If a packet enters a network segment that does not support the DSCP value,
this could result in the connection not receiving the network treatment
it expects. The DSCP value in this packet could also be remarked as the
packet travels along the network path, changing the requested treatment.

When multiplexing multiple flows over a QUIC connection, the selected
DSCP value should be the one associated with the highest priority requested
for all multiplexed flows.

```

draft-ietf-quic-manageability.md
```diff
@@ -1055,18 +1055,28 @@ It is expected that any QoS handling in the network, e.g. based on use of
...
 network paths or any other non-uniform treatment of packets belong to the same
 connection could result in variations in order, delivery rate, and drop rate.
 As feedback about loss or delay of each packet is used as input to
 the congestion controller, these variations could adversely affect performance.
-
 Depending on the loss recovery mechanism implemented, QUIC may be
 more tolerant of packet re-ordering than traditional TCP traffic (see
-{{packetnumber}}). However, it cannot be known by the network which exact
-recovery mechanism is used and therefore reordering tolerance should be
+{{packetnumber}}). However, the recovery mechanism used by a flow cannot be
```
mirjak, line 1067:
```suggestion
{{packetnumber}}). However, the recovery mechanism used by a flow cannot be
```

draft-ietf-quic-manageability.md
```diff
@@ -1055,18 +1055,28 @@ It is expected that any QoS handling in the network, e.g. based on use of
...
+known by the network and therefore reordering tolerance should be
 considered as unknown.
 
+If a packet enters a network segment that does not support the DSCP value,
```
britram, line 1071: This text would be a better fit for the applicability draft, no?

britram (changes requested):

draft-ietf-quic-manageability.md
gorryfair: OK - I see we need two PRs. I'll edit this one as NiTs for manageability, and start another for applicability!

draft-ietf-quic-manageability.md
```diff
@@ -1055,16 +1055,17 @@ It is expected that any QoS handling in the network, e.g. based on use of
 DiffServ Code Points (DSCPs) {{?RFC2475}} as well as Equal-Cost
 Multi-Path (ECMP) routing, is applied on a per flow-basis (and not per-packet)
```
gloinul, line 1056: I would suggest making it clear that this is per transport flow, i.e. and not streams or their datagram equivalent that is called datagram flows. So maybe it would be clearer to state this in relation ship to a QUIC connection?

draft-ietf-quic-manageability.md
```diff
@@ -1055,16 +1055,17 @@ It is expected that any QoS handling in the network, e.g. based on use of
...
 and as such that all packets belonging to the same QUIC connection get uniform
-treatment. Using ECMP to distribute packets from a single flow across multiple
+treatment.
```
gloinul, line 1058: I wonder if a reference to RFC7657 would make sense also here.

gloinul (commented):
I think the recommendation of using a single DSCP value per QUIC connection transport path at any time in the connections life time could be clearer. I also think that adding a reference to RFC 7657 could help to provide background why that is necessary without having to expand this text.

draft-ietf-quic-manageability.md
gorryfair: Actually connection"lifetime" is too prescriptive... if you really want/need a sender can change the DSCP within a connection, but this really needs to be infrequent to prevent significant disruption to the CC, replay-protection, etc. I'd rather not drill there. I elsewhere added a ref to an RFC that says all this.

gorryfair: /flow/ is used rather a lot in this spec. I think it might be clear, or require a change-all to /network flow/ to avoid potential misreading with QUIC datagram flows? Thoughts?

gloinul: To be clear I didn't mean the full life time of the QUIC connection, I do mean use the same DSCP over a consecutive set of packets and one should really not switch back and forth. So I think we are in agreement. 

Considering that the QUIC spec do have recommendations for path changes, one way to deal with this that requires little text is to basically to treat any change of DSCP as a path when it comes to CC and recovery and basically execute Section 9.4 of RFC 9000. However, I do realize that likely should be speced better as you also have to consider what the validation is for a DSCP change and if there is a risk to use DSCP changes for DoS attacks keeping down the rate. So you are likely correct that keeping this simple and only be clearer on that changes should be very infrequent.

britram: @gorryfair could you come to a conclusion here so we can merge? thanks!

draft-ietf-quic-manageability.md
britram: PRs cheerfully accepted :)

britram: Actually, scratch that; I think the reference belongs in Applicability (PR #383).

draft-ietf-quic-manageability.md
britram: Actually, the resolution to this discussion also looks like it probably belongs in the Applicability draft (PR #383); will merge this and you can continue the discussion over there :)

PR #380, opened 2021-05-26
Note DCID must be extracted from Client Initial

britram:
Addresses #344

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -647,7 +647,7 @@ gradually disappear over time.
 When the version has been identified as QUIC version 1, the packet type needs to
 be verified as an Initial packet by checking that the third and fourth bits of
 the header are both set to 0. Then the Destination Connection ID needs to be
-extracted from the Client Initial to calculate the Initial secret using the
+extracted from the packet. The Initial secret is calculated using the
 version-specific Initial salt, as described in {{Section 5.2 of QUIC-TLS}}.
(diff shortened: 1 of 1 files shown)


martinduke (approved):

draft-ietf-quic-manageability.md
```diff
@@ -647,10 +647,16 @@ gradually disappear over time.
 When the version has been identified as QUIC version 1, the packet type needs to
 be verified as an Initial packet by checking that the third and fourth bits of
 the header are both set to 0. Then the Destination Connection ID needs to be
-extracted to calculate the Initial secret using the version-specific Initial
-salt, as described in {{Section 5.2 of QUIC-TLS}}. The length of the connection
-ID is indicated in the 6th byte of the header followed by the connection ID
-itself.
+extracted from the Client Initial to calculate the Initial secret using the
```
martinthomson, line 650:
```suggestion
extracted from the packet. The Initial secret is calculated using the
```

"Client Initial" is not really a thing in this context (you don't really want to reuse that context here, I think).  As you are talking about packets in the abstract, I think that this is right.  The next paragraph covers the important gotcha.

martinthomson (commented):
Much better.  My comment here is mostly a nit.  I think that you are implicitly talking about the first packet from the client, so maybe you can avoid using the "Client Initial" thing here (if you are using the abstract concept you introduced), "client Initial" (that is, Initial packets sent by a client), or anything that might be confusing.

PR #383, opened 2021-05-26
Proposed DSCP text for Applications usage

gorryfair:
Separates  #379 for manageability and this PR for applicability

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -770,16 +770,24 @@ however, the use of more modern cryptographic algorithms is highly recommended.
 
 # Quality of Service (QoS) and DSCP
 
-QUIC assumes that all packets of a QUIC connection, or at least with the
+QUIC, as defined in {{!RFC9000}}, has a single congestion controller and 
+recovery handler. This design 
+assumes that all packets of a QUIC connection, or at least with the
 same 5-tuple {dest addr, source addr, protocol, dest port, source port}, will
 receive similar network treatment since feedback about loss or delay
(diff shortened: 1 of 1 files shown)


gloinul:
I actually find this sentence weak: 

"Therefore it is not recommended to use different DiffServ Code Points (DSCPs) {{?RFC2475}} for ...

Considering that QUIC v1 as defined has a single congestion controller and recovery handler and don't treat different DSCP values as separate transport flows I would think that this should be stronger formulated. 

However, I could see that one could provide a bit more context to this DSCP and single QUIC connections vs multiple QUIC streams discussion by referencing Section 5.1 of https://datatracker.ietf.org/doc/rfc7657/ and be explicit about the point that QUIC v1 is similar to SCTP in regards to the issues raised.

draft-ietf-quic-applicability.md
```diff
@@ -770,17 +770,34 @@ however, the use of more modern cryptographic algorithms is highly recommended.
 
 # Quality of Service (QoS) and DSCP
 
-QUIC assumes that all packets of a QUIC connection, or at least with the
+QUIC, as defined in {{!RFC9000}}, has a single congestion controller and 
+recovery handler. This design 
+assumes that all packets of a QUIC connection, or at least with the
 same 5-tuple {dest addr, source addr, protocol, dest port, source port}, will
```
gloinul, line 776: Shouldn't actually DSCP be part of the tuple?

gloinul (commented):
I think the new text do work.

PR #385, opened 2021-05-27
PR to refer to network-layer management/analysis

gorryfair:
Refers to quicwg#362. 
This also addresses /extremely limited/ ECN visibility (limited for sure, but we don't need to say "extremely").
Fixed multiple definitions of RTT
Refered to metrics, etc for encrypted traffic wrt network-layer measurement via a REF.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -833,7 +833,9 @@ packet header (Section 6 of
 {{?I-D.ietf-tsvwg-transport-encrypt}} describes use of operations,
 administration and management (OAM) information).
 Using network-layer approaches also has the advantage that common observation
-and analysis tools can be consistently used by multiple transport protocols.
+and analysis tools can be consistently used by multiple transport protocols,
+however, these techniques are often limited to measurements within one or
+multiple cooperating domains.
 
 ## Stateful Treatment of QUIC Traffic {#sec-stateful}
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -721,7 +721,7 @@ traffic is protected and ACKs may be padded, padding is not required.
 
 ## Round-Trip Time (RTT) Measurement {#sec-rtt}
 
-The round-trip time of QUIC flows can be inferred by observation once per flow,
+The round-trip time (RTT) of QUIC flows can be inferred by observation once per flow,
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -823,9 +824,17 @@ techniques and how QUIC's design impacts them.
...
+wire image. Limited observation of upstream congestion may be
 possible via the observation of CE markings on ECN-enabled QUIC traffic.
 
+On-path devices can also make measurements of RTT, loss and other
+performance metrics when information is carried in an additional network-layer
+packet header (Section 6 of
+{{?I-D.ietf-tsvwg-transport-encrypt}} describes use of operations,
+administration and management (OAM) information).
+Using network-layer approaches also has the potential that common observation
+and analysis tools can be consistently used by multiple transport protocols.
```
mirjak, line 836:
```suggestion
and analysis tools can be consistently used by multiple transport protocols,
however, these techniques are often limited to measurements within one or
multiple cooperating domains.
```
Proposing this change as it otherwise sounds that we don't need any in-band e2e measurement support which I don't think is true or intended. However, let me know if what I wrote is actually correct...?

draft-ietf-quic-manageability.md
```diff
@@ -823,9 +824,17 @@ techniques and how QUIC's design impacts them.
...
-wire image. Extremely limited observation of upstream congestion may be
+wire image. Limited observation of upstream congestion may be
 possible via the observation of CE markings on ECN-enabled QUIC traffic.
 
+On-path devices can also make measurements of RTT, loss and other
+performance metrics when information is carried in an additional network-layer
+packet header (Section 6 of
+{{?I-D.ietf-tsvwg-transport-encrypt}} describes use of operations,
+administration and management (OAM) information).
+Using network-layer approaches also has the potential that common observation
```
ianswett, line 835:
```suggestion
Using network-layer approaches also has the advantage that common observation
```

draft-ietf-quic-manageability.md
ianswett: I think this is a reasonable addition, but I'd prefer a separate sentence starting with "However, "

gorryfair:
Thanks for the suggestions from Ian and Mirjay, they were of course incorporated.

mirjak (approved):

Issue #352, opened 2021-05-10
Clarify coalescing and multiple packets in "The QUIC Handshake" section


ianswett:
QUIC handshakes may or may not coalesce packets, which is stated, but I'm worried it could be missed by a reader.

I think it'd be good to clarify that any portion of the server or client flight could span multiple datagrams.

In particular, I'm worried this paragraph could mislead people, because unless the connection is resumed, the Handshake CRYPTO data typically spans multiple datagrams.  Similarly, the client or server's Initial can span multiple datagrams, though that's not currently common:
> A typical handshake starts with the client sending of a Client Initial datagram as shown in {{fig-client-initial}}, which elicits a Server Initial datagram as shown in {{fig-server-initial}} typically containing three packets: an Initial packet with the Server Initial, a Handshake packet with the rest of the server's side of the TLS handshake, and initial 1-RTT data, if present."

mirjak:
@ianswett can you maybe propose a PR? :-)

ianswett:
Yes, possibly later today after I finish reviewing the draft.

mirjak:
@ianswett are you still planning to do a PR here?

ianswett:
I will unless someone beats me to it, but I've been on paternity leave for 10 days.  @DavidSchinazi had some suggested text for the SNI section, but I can try to fix "The QUIC Handshake" and assume David's text is a good start for SNI.

ianswett:
While writing the PR, I discovered a problem I'm not sure how to fix, which is how does an observer know when the handshake is complete for the server, given the server can start using 1-RTT keys immediately after receiving the client hello?

Receipt of the client's 1-RTT packet doesn't guarantee anything, because the server can't process it if it doesn't have keys.

PR #387, opened 2021-06-02
Clarify "The QUIC Handshake"

ianswett:
As noted in #352 the "The QUIC Handshake" is written in a way that could cause readers to assume things such as there's only one datagram containing a Client Initial.

This PR changes the naming from 'Client/Server Initial/Finished datagram' to 'Client/Server Initial/Finished flight' and makes related changes to make it read correctly.  The diagrams have not been fully updated yet.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -305,8 +305,8 @@ per-version constant and considered visible in this illustration. The content
 of QUIC Handshake packets are encrypted using keys established during the
 initial handshake exchange, and are therefore not visible.
 
-Initial, Handshake, and 1-RTT packets belong to different cryptographic and transport
-contexts. The Client Completion {{fig-init-complete}} and the
+Initial, Handshake, and 1-RTT packets belong to different cryptographic and
+transport contexts. The Client Completion {{fig-init-complete}} and the
 Server Completion {{fig-hs-complete}} flights conclude the Initial and Handshake
 contexts, by sending final acknowledgments and CRYPTO frames.
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -252,7 +252,7 @@ contexts ("Initial", which contains observable payload, and "Handshake" and
 "1-RTT", which do not). QUIC packets in separate contexts during the handshake
 can be coalesced (see {{coalesced}}) in order to reduce the number of UDP
 datagrams sent during the handshake.  QUIC packets can be lost and reordered,
-so packets within a flight may not be sent close in time, though the sequence
+so packets within a flight might not be sent close in time, though the sequence
 of the flights will not change, because one flight depends upon the peer's
 previous flight.
 
@@ -285,7 +285,7 @@ three types of packets: Initial packet(s) with the beginning of the server's
 side of the TLS handshake, Handshake packet(s) with the rest of the server's
 portion of the TLS handshake, and 1-RTT packet(s), if present.
 
-The Client Completion packets contain at least one Handshake packet and
+The Client Completion flight contains at least one Handshake packet and
 could also include an Initial packet.
 
 Datagrams that contain an Initial packet (Client Initial, Server
@@ -305,9 +305,9 @@ per-version constant and considered visible in this illustration. The content
 of QUIC Handshake packets are encrypted using keys established during the
 initial handshake exchange, and are therefore not visible.
 
-Initial, Handshake, and 1-RTT packets belong to cryptographic and transport
+Initial, Handshake, and 1-RTT packets belong to different cryptographic and transport
 contexts. The Client Completion {{fig-init-complete}} and the
-Server Completion {{fig-hs-complete}} flights conclude these first two
+Server Completion {{fig-hs-complete}} flights conclude the Initial and Handshake
 contexts, by sending final acknowledgments and CRYPTO frames.
 
 ~~~~~
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -353,7 +353,7 @@ QUIC Initial packets.
 | 1-RTT encrypted payload                                    |
 +------------------------------------------------------------+
 ~~~~~
-{: #fig-server-initial title="Typical Server Initial datagram pattern"}
+{: #fig-server-initial title="Coalesced Server Initial datagram pattern"}
 
 The Server Initial datagram also exposes version number, source and destination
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -299,38 +301,36 @@ packet for QUIC to be used.
...
 
-Initial, Handshake, and the 1-RTT packets transmitted after the handshake
-belong to cryptographic and transport contexts. The Client Completion
-{{fig-init-complete}} and the Server Completion {{fig-hs-complete}} datagrams
-finish these first two contexts, by sending the final acknowledgment and
-finishing the transmission of CRYPTO frames.
+Initial, Handshake, and 1-RTT packets belong to cryptographic and transport
+contexts. The Client Completion {{fig-init-complete}} and the
+Server Completion {{fig-hs-complete}} flights conclude these first two
+contexts, by sending final acknowledgments and CRYPTO frames.
```
martinthomson, line 311:
```suggestion
Server Completion {{fig-hs-complete}} flights conclude the Initial and Handshake
contexts, by sending final acknowledgments and CRYPTO frames.
```

If you look at this, it's not a great match to reality.  The Initial context concludes at the server when it sends its first Handshake packet.  Is this "lies to children" and therefore acceptable?

draft-ietf-quic-manageability.md
```diff
@@ -425,12 +426,11 @@ datagram can also carry 0-RTT data, as shown in {{fig-client-initial-0rtt}}.
...
    title="Typical 0-RTT Client Initial datagram pattern"}
 
-In a 0-RTT Client Initial datagram, the PADDING frame is only present if
-necessary to increase the size of the datagram with 0-RTT data to at least 1200
-bytes. Additional datagrams containing only 0-RTT packets with long headers
-could be sent from the client to the server after the Client Initial datagram,
-containing the rest of the 0-RTT data. The amount of 0-RTT protected data
-that can be sent in the first flight is limited by the initial congestion
+When a 0-RTT packet is coalesced with an Initial packet, the datagram
+will be padded to 1200 byes. Additional datagrams containing only 0-RTT
```
martinthomson, line 430: Note: This is also "lies to children".  If a Handshake packet is coalesced with these (which can happen, depending on how the client works; ours will do this in some circumstances), then the datagram doesn't need to be padded.

martinthomson (approved):
This is a good improvement.

Don't pay too much attention to my nitpicking.  I love finding ways in which the descriptions of QUIC can be wrong.  This draft necessarily simplifies, which means it is highly prone to inaccuracies of the sort I point out here.  I don't think that these are fatal problems; feel free to ignore them.  Or, feel free to apply weasel words like "might" or "often" as necessary.  I might be inclined to contest those though; see the last comment: our stack could generate the datagram I describe up to 60% of the time.

draft-ietf-quic-manageability.md
ianswett: Yeah, this was my best effort at rewriting what was there to be clearer and more accurate, but it's still a bit misleading or possibly just not helpful.

I'm not sure if the intent is to detail when no more packets from an encryption level will be sent, because the keys have been discarded?

draft-ietf-quic-manageability.md
ianswett: It doesn't?  I'm missing that in: https://www.rfc-editor.org/rfc/rfc9000.html#name-initial-datagram-size

I agree that in practice, if the client has Handshake keys but hasn't used them(ie: it's sending an Initial ACK before the Handshake packet), then there's no reason to pad the datagram, but I can't find text which makes an exception for that.

ianswett (commented):
Thanks for the suggestions Martin, very helpful.

draft-ietf-quic-manageability.md
```diff
@@ -405,7 +405,8 @@ information; observing it serves only to determine that the handshake has
 completed.
```
ianswett, line 405: I'm actually not sure how one determines the handshake is complete from the server completion.  How does one tell the difference between the server sending 0.5RTT and 1RTT containing HANDSHAKE_DONE?

draft-ietf-quic-manageability.md
martinthomson: You know, I don't think that this is in the spec.  It's not a condition that the spec considers.  So maybe I can take that back.

It's what our implementation does though.  There is no point padding if you are sending Handshake, because the Handshake packet will unblock the server anti-amplification limit and you know that the server can process the Handshake packet (the only reason to send Handshake in this case is to acknowledge a Handshake packet).

Some implementations won't even send Initial in this case because they drop the keys.  We have this extra wrinkle that allows us to acknowledge Initial packets as a final gasp before dropping the keys.  It's a courtesy only and not reliable, but it means that we send Initial + Handshake + 0-RTT all in the same datagram.  AND that datagram won't be padded, even though it contains an Initial.

So I'm totally OK with this text remaining.  I certainly don't want to have to explain all of that.

draft-ietf-quic-manageability.md
martinthomson: Maybe this is worth a note:
> The Server Completion flight might not be observable as it is indistinguishable from packets the server sends after the Server Initial flight.

draft-ietf-quic-manageability.md
ianswett: The 'courtesy ACKs' are actually useful because they provide an extra RTT sample early in the connection, so thanks for your courtesy :)

draft-ietf-quic-manageability.md
```diff
@@ -299,38 +301,36 @@ packet for QUIC to be used.
...
 ~~~~~
 +----------------------------------------------------------+
 | UDP header (source and destination UDP ports)            |
 +----------------------------------------------------------+
 | QUIC long header (type = Initial, Version, DCID, SCID) (Length)
 +----------------------------------------------------------+  |
-| QUIC CRYPTO frame header                                 |  |
+| QUIC Frames (ie: CRYPTO, PADDING, PING)                  |  |
 +----------------------------------------------------------+  |
 | TLS Client Hello (incl. TLS SNI)                         |  |
```
britram, line 321: does the TLS Client Hello not live within a QUIC frame? this makes it look otherwise in my reading.

ianswett: Thanks, I updated it to be clearer.  I also reverted part of my change and just put "Example" instead of the former "Typical", because the number of ways a client hello can be sequenced is extraordinarily large.

DavidSchinazi (approved):

PR #388, opened 2021-06-30
SNI parsing proposals by David

### LINKED ISSUE #358: Manageability: SNI extraction assumes a single CRYPTO frame
The [SNI extraction section of the manageability draft](https://datatracker.ietf.org/doc/html/draft-ietf-quic-manageability-11#section-3.4.1) mentions `parsing the QUIC CRYPTO Frame containing the TLS ClientHello`. This makes the assumption that INITIAL packets carry a single CRYPTO frame. That assumption might not always hold in modern browsers, so the text should clarify that any middlebox that wishes to extract the SNI MUST perform reconstruction of the crypto stream from one or more CRYPTO frames in one or more INITIAL packets. The current assumption might lead middlebox vendors to write deploy software that doesn't work if there are multiple CRYPTO frames. The section correctly states `There might be multiple CRYPTO frames` but that comes late in the section and doesn't detail what middlebox vendors should do about it.

That section also contains another incorrect assumption: `In QUIC version 1, the packet is expected to contain only CRYPTO frames and optionally PADDING frames`. There are other frame types that are legal in INITIAL packets and may be sent by browsers.

mirjak:
fixes #358

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -633,9 +633,14 @@ and other protocols that use TLS.
 
 ### Extracting Server Name Indication (SNI) Information
 
+
+
+
+
+
 If the ClientHello is not encrypted, SNI can be derived from the client's
 Initial packet by calculating the Initial secret to decrypt the packet payload
-and parsing the QUIC CRYPTO Frame containing the TLS ClientHello.
+and parsing the QUIC CRYPTO frame(s) containing the TLS ClientHello.
 
 As both the derivation of the Initial secret and the structure of the Initial
 packet itself are version-specific, the first step is always to parse the
@@ -661,7 +666,6 @@ other than the one used to generate the Initial secret. Therefore, attempts to
 decrypt these packets using the procedure above might fail unless the Initial
 secret is retained by the observer.
 
-
 To determine the end of the header and find the start of the payload, the packet
 number length, the source connection ID length, and the token length need to be
 extracted. The packet number length is defined by the seventh and eight bits of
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -673,14 +677,16 @@ specified in {{Section 16 of QUIC-TRANSPORT}}.
...
-consisting of a single zero byte, may occur before, after, or between CRYPTO
-frames. There might be multiple CRYPTO frames.  Finally, an extension might
-define additional frame types which could be present.
+to TLS over TCP connections. Note that there can be multiple CRYPTO frames,
+and they might not be in order, so reassembling the CRYPTO stream by parsing
+offsets and lengths is required. Further, the client's Initial packet may contain
+other frames, so the first bytes of each frame need to be checked to identify the
+frame type, and if needed skipped over it. Note that the length of the frames is
+dependent on the frame type; see {{Section 18 of QUIC-TRANSPORT}}. E.g. PADDING
+frames, each consisting of a single zero byte, may occur before, after, or
```
mirjak, line 686:
```suggestion
offsets and lengths is required. Further, the client's Initial packet may
contain other frames, so the first bytes of each frame need to be checked
to identify the frame type, and if needed skipped over it. Note that the
length of the frames is dependent on the frame type; see {{Section 18 of
QUIC-TRANSPORT}}. E.g. PADDING frames, each consisting
of a single zero byte, may occur before, after, or
```

draft-ietf-quic-manageability.md
```diff
@@ -633,9 +633,14 @@ and other protocols that use TLS.
 
 ### Extracting Server Name Indication (SNI) Information
 
+
+
+
+
+
```
mirjak, line 640:
```suggestion
```

ianswett:
I think there are still a few potentially misleading statements left in the text.  I'll send out a PR.

PR #389, opened 2021-06-30
Tweaks to the Extracting SNI section

ianswett:
Follow-on to #388


mirjak:
Thanks @ianswett! As there are clarifications/editorial, I think we will wait for the AD review and take this PR up with a next revision before or after IETF last call. But I think it's ready to merge!

ianswett:
You're welcome.  Agreed they're editorial, so merge whenever it makes sense.

PR #390, opened 2021-07-02
Add recent contributors to managability

ianswett:
I found 5 people who landed PRs since this list was last updated.

I added them in an arbitrary order, feel free to suggest changes.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -1190,6 +1190,11 @@ The following people have contributed text to sections of this document:
 * David Schinazi
 * Gorry Fairhurst
 * Chris Box
+* Matt Joras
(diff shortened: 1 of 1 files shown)


mirjak:
Not sure this merge was correct. I don't think we listed everybody with a PR necessarily here rather than in the ACK part below. If it was only nits or editorial changes we didn't list them as contributors. Also if we change someone to contributor we should remove them from the ACK section.

ianswett:
That makes sense MIrja, I'll let you and Brian discuss this and decide what to do if that SG?

mirjak:
Yes, we will fix this (in the sense of making is consistent). It's definitely good to review and double-check we have all the right people in with the next version!

PR #392, opened 2021-07-19
Talk about source ports

mnot:
as discussed on-list

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -662,7 +662,7 @@ some instances, might cause the client to avoid using QUIC for that service for
 a period of time, downgrading to a non-UDP protocol (see {{fallback}}).
 
 As a result, client implementations are encouraged to avoid using source ports
-associated with protocols known to be vulnerable to reflection attacks.
+associated with protocols known to be vulnerable to reflection attacks. Note that the list above is only indicative; other source ports might be considered reflection vectors as well.
 
 
 # Connection Migration
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -647,7 +647,7 @@ Some UDP protocols are vulnerable to reflection attacks, where an attacker is ab
 
 * port 53 - DNS {{?RFC1034}}
 * port 123 - NTP {{?RFC5905}}
-* port 1900 - SSDP {{?SSDP}}
+* port 1900 - SSDP {{SSDP}}
 * port 5353 - mDNS {{?RFC6762}}
 * port 11211 - memcached
 
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -654,9 +654,9 @@ reflection attacks, often due to server misconfiguration:
 * port 5353 - mDNS {{?RFC6762}}
 * port 11211 - memcached
 
-Services might block source ports of protocols known to be vulnerable to
-reflection, to avoid the overhead of processing large numbers of packets by
-their QUIC implementations. However, this practice has negative effects on
+Services might block source ports associated with protocols known to be
+vulnerable to reflection attacks, to avoid the overhead of processing large
+numbers of packets. However, this practice has negative effects on
 clients: not only does it require establishment of a new connection, but in
 some instances, might cause the client to avoid using QUIC for that service for
 a period of time, downgrading to a non-UDP protocol (see {{fallback}}).
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -662,7 +662,9 @@ some instances, might cause the client to avoid using QUIC for that service for
 a period of time, downgrading to a non-UDP protocol (see {{fallback}}).
 
 As a result, client implementations are encouraged to avoid using source ports
-associated with protocols known to be vulnerable to reflection attacks. Note that the list above is only indicative; other source ports might be considered reflection vectors as well.
+associated with protocols known to be vulnerable to reflection attacks. Note
+that the list above is only indicative; other source ports might be considered
+reflection vectors as well.
 
 
 # Connection Migration
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -643,7 +643,10 @@ protocols, like HTTP/3.
 
 ## Source Port Selection
 
-Some UDP protocols are vulnerable to reflection attacks, where an attacker is able to direct traffic to a third party as a denial of service. For example, these source ports are associated with applications known to be vulnerable to reflection attacks (often due to server misconfiguration):
+Some UDP protocols are vulnerable to reflection attacks, where an attacker is
+able to direct traffic to a third party as a denial of service. For example,
+these source ports are associated with applications known to be vulnerable to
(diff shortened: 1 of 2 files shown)


martinthomson (commented):
This needs to talk about the possibility that a NAT could rewrite a source port to use one of these values.  Clients cannot control this and so can't guarantee that these ports will be avoided in all cases.  In other words, this is fragile and that needs to be acknowledged.

draft-ietf-quic-applicability.md
```diff
@@ -620,6 +641,29 @@ consider how the protocol will manage different QUIC versions.
...
+* port 1900 - SSDP {{SSDP}}
+* port 5353 - mDNS {{?RFC6762}}
+* port 11211 - memcached
+
+Services might block source ports of protocols known to be vulnerable to
+reflection, to avoid the overhead of processing large numbers of packets by
+their QUIC implementations. However, this practice has negative effects on
+clients: not only does it require establishment of a new connection, but in
+some instances, might cause the client to avoid using QUIC for that service for
+a period of time, downgrading to a non-UDP protocol.
```
mirjak, line 662:
```suggestion
a period of time, downgrading to a non-UDP protocol (see {{fallback}}).
```

mnot:
@martinthomson if we want to reach NAT vendors, maybe that text should be in the manageability doc?

enygren:
Should we recommend that using UDP source ports 10k-64k is preferable where possible (with some exceptions like 11211) and that NATs may need to take care to avoid filtered ports?  This seems like it will be a fragile topic on an ongoing basis as the set of filtered ports will grow over time, but it so far seems less likely for them to be over 10k or even over 32k.

Digging into a set of UDP ports receiving the most scans against a widely deployed firewall sees some of these on top:
* 137
* 500
* 53
* 123
* 1900
* 161
* 389
* 3702
* 5353
* 1434
* 5683
* 1434
* 5683
* 1484
* 5355
* 11211
* 6881
* 1194
* 111
* ...

Fetch maintains a list of problematic *destination* ports, but it is fairly TCP-specific.  Agreed that we'd preferably have some better way to maintain this list outside of QUIC.

mnot:
@enygren this isn't about commonly-scanned ports; it's about those specifically used for reflection attacks.

enygren:
@mnot Agreed, that is just a datapoint as what matters is what people put filters on in firewalls or routers.  Those used for reflection attacks are top ones to avoid, but ports that are widely scanned also seem worth avoiding.   (In "scanned" here I'm not making a distinction between "scanned" and "target of a reflection attack")

mnot:
@enygren the ports you list are destination ports, I assume.

enygren:
Yes, the ones I list are destination ports (which could/will be network-observed source ports for any protocols which reflect).

martinthomson:
As the primary concern is volume of traffic, the narrow formulation from Mark is my preference. If any of the other ports can be shown to be a reflection vector, let us talk about adding them. QUIC is pretty robust in terms of managing junk packets, so making this about volume will keep it focused.

draft-ietf-quic-manageability.md
```diff
@@ -991,6 +991,10 @@ While QUIC endpoints are often able to survive address changes, e.g. by NAT
 rebindings, blocking a portion of the traffic based on 5-tuple hashing increases
 the risk of black-holing an active connection when the address changes.
 
+Note that some source ports are assumed to be reflection attack vectors by some
+servers; see {{Section 8.1 of I-D.ietf-quic-applicability}}. As a result, NAT
+rebinding to these source ports can result in that traffic being blocked.
```
mirjak, line 996: Is this only about rebinding or any port selection of NATs?

mnot: I th ink it's mostly about rebinding.

martinthomson: I think that it is just "binding" rather than "rebinding".  I read "rebinding" as what happens when flow state is lost and a new port binding is assigned to that flow.  But this is just "binding", which is any time a port is bound to a flow.

draft-ietf-quic-applicability.md
```diff
@@ -620,6 +641,31 @@ consider how the protocol will manage different QUIC versions.
...
+Services might block source ports of protocols known to be vulnerable to
+reflection, to avoid the overhead of processing large numbers of packets by
+their QUIC implementations. However, this practice has negative effects on
+clients: not only does it require establishment of a new connection, but in
+some instances, might cause the client to avoid using QUIC for that service for
+a period of time, downgrading to a non-UDP protocol (see {{fallback}}).
+
+As a result, client implementations are encouraged to avoid using source ports
+associated with protocols known to be vulnerable to reflection attacks. Note
+that the list above is only indicative; other source ports might be considered
```
ianswett, line 666:
```suggestion
that the list above is not exhaustive; other source ports might be considered
```

ianswett (commented):
Looks sensible, some suggestions.

draft-ietf-quic-manageability.md
mnot: A quick look around shows a lot more consistent use of 'rebinding' for this case; e.g., [here](https://tools.ietf.org/id/draft-duke-quic-natsupp-00.html)

draft-ietf-quic-applicability.md
```diff
@@ -620,6 +641,31 @@ consider how the protocol will manage different QUIC versions.
...
+
+* port 53 - DNS {{?RFC1034}}
+* port 123 - NTP {{?RFC5905}}
+* port 1900 - SSDP {{SSDP}}
+* port 5353 - mDNS {{?RFC6762}}
+* port 11211 - memcached
+
+Services might block source ports of protocols known to be vulnerable to
+reflection, to avoid the overhead of processing large numbers of packets by
+their QUIC implementations. However, this practice has negative effects on
```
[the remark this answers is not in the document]
mnot: I don't think it's the port that's vulnerable to attacks; it's the underlying protocol.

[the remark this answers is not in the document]
britram:
```suggestion
Services might block source ports associated with protocols known to be
vulnerable to reflection attacks, to avoid the overhead of processing large
numbers of packets. However, this practice has negative effects on
```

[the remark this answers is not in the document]
britram: split the difference between Ian's and Mark's suggestions?

britram (approved):

draft-ietf-quic-manageability.md
britram:
```suggestion
binding to these source ports can result in that traffic being blocked.
```

draft-ietf-quic-manageability.md
```diff
@@ -959,7 +959,7 @@ Admitting a few packets allows the QUIC endpoint to determine that the path
 accepts QUIC. Sudden drops afterwards will result in slow and costly timeouts
 before abandoning the connection.
 
-## UDP Blocking or Throttling {#sec-udp-1312}
+## UDP Blocking, Throttling, and Rebinding {#sec-udp-1312}
```
britram, line 962:
```suggestion
## UDP Blocking, Throttling, and NAT Binding {#sec-udp-1312}
```

PR #424, opened 2021-11-24
Use of encrypted message for Version negotiation

zaheduzzaman:
Currently the document uses "will".  This might be a very valid assumption, but still it is an assumption hence I think it is better to use "May" instead of "Will".

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -493,7 +493,7 @@ to on-path observers.
 Version Negotiation packets are used by the server to indicate that a requested
 version from the client is not supported (see {{Section 6 of QUIC-TRANSPORT}}.
 Version Negotiation packets are not intrinsically protected, but future QUIC
-versions may use later encrypted messages to verify that they were authentic.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -493,7 +493,7 @@ to on-path observers.
 Version Negotiation packets are used by the server to indicate that a requested
 version from the client is not supported (see {{Section 6 of QUIC-TRANSPORT}}.
 Version Negotiation packets are not intrinsically protected, but future QUIC
-versions will use later encrypted messages to verify that they were authentic.
+versions may use later encrypted messages to verify that they were authentic.
```
martinthomson, line 496:
```suggestion
versions could use later encrypted messages to verify that they were authentic.
```

We have evidence that they will though: draft-ietf-quic-version-negotiation, so I'm not sure that this is something that we need to prevaricate on.

zaheduzzaman: that is even better, we can tell it more substantiate way..

PR #432, opened 2021-12-13
ref to RFC9065 and clarification about troubleshooting

### LINKED ISSUE #410: Clarification on troubleshooting and in-network modification is needed in the introduction
https://github.com/quicwg/ops-drafts/blob/3893028c41567d89a6f416049cff901ce87a5c21/draft-ietf-quic-manageability.md?plain=1#L88, I am not sure I get what this paragraph is supposed to mean. Number of things -

- first of all it is not clear that integrity protection actually simply the troubleshooting without defining that kind of troubleshooting we are referring to here and who is doing those, even though it is true that now network nodes cannot modify the transport layer information.
- we are making a statement that in-network operations depends on modification of data, what is the data here? payload? header information? both? I think here we can reference to the I-D.ietf-tsvwg-transport-encrypt to setup the background information required to parse the statement correctly.
- I am not sure how a proxy here can support modification of data. Yes, that is my read of this last of the statement. Hence, hence we might want to clarify what is the scope of this collaboration that we are referring to .

mirjak:
fixes #410

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -85,12 +85,13 @@ achieved through integrity protection of the wire image {{?WIRE-IMAGE=RFC8546}}.
 Encryption of most control signaling means that less information is visible to
 the network than is the case with TCP.
 
-Integrity protection can also simplify troubleshooting, because none of the
-nodes on the network path can modify transport layer information. However, it
-means in-network operations that depend on modification of data are not
-possible without the cooperation of an QUIC endpoint. This might be possible
-with the introduction of a proxy which authenticates as an endpoint.
-Proxy operations are not in scope for this document.
+Integrity protection can also simplify troubleshooting at the end points as none
+of the nodes on the network path can modify transport layer information. However,
+it means in-network operations that depend on modification of data (e.g. see 
+{{?RFC9065}} are not possible without the cooperation of an QUIC endpoint. Such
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -85,12 +85,13 @@ achieved through integrity protection of the wire image {{?WIRE-IMAGE=RFC8546}}.
...
-possible without the cooperation of an QUIC endpoint. This might be possible
-with the introduction of a proxy which authenticates as an endpoint.
-Proxy operations are not in scope for this document.
+Integrity protection can also simplify troubleshooting at the end points as none
+of the nodes on the network path can modify transport layer information. However,
+it means in-network operations that depend on modification of data (e.g. see 
+{{?RFC9065}} are not possible without the cooperation of an QUIC endpoint. Such
+cooperation might be possible with the introduction of a proxy which
+authenticates as an endpoint. Proxy operations are not in scope for this
+document.
```
mirjak, line 94:
```suggestion
of the nodes on the network path can modify transport layer information.
However, it means in-network operations that depend on modification of data
(e.g. see {{?RFC9065}} are not possible without the cooperation of an QUIC
endpoint. Such cooperation might be possible with the introduction of a proxy
which authenticates as an endpoint. Proxy operations are not in scope for this
document.
```

PR #446, opened 2022-01-11
Quic might change flow ids

### LINKED ISSUE #422: How does conection migration impacts the per-flow basis flow treatment?
Section 4.8 of manageability draft expects that a per-flow basis DSCP and ECMP routing policy is used to ensure same treatment for all the packets belonging to same QUIC connection. I was wondering how does path migration impacts this? As the new path ,potentially via another network, has no idea about QUIC connection insuring all packets belonging to the same QUIC connection get uniform treatment will be difficult. what am I missing here?

mirjak:
fixes #422

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -1069,7 +1069,7 @@ Retry packets on behalf of consenting servers.
 It is expected that any QoS handling in the network, e.g. based on use of
 DiffServ Code Points (DSCPs) {{?RFC2475}} as well as Equal-Cost
 Multi-Path (ECMP) routing, is applied on a per flow-basis (and not per-packet)
-and as such that all packets belonging to the same active QUIC connectio
+and as such that all packets belonging to the same active QUIC connection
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -1069,8 +1069,8 @@ Retry packets on behalf of consenting servers.
 It is expected that any QoS handling in the network, e.g. based on use of
 DiffServ Code Points (DSCPs) {{?RFC2475}} as well as Equal-Cost
 Multi-Path (ECMP) routing, is applied on a per flow-basis (and not per-packet)
-and as such that all packets belonging to the same QUIC connection get uniform
-treatment.
+and as such that all packets belonging to the same active QUIC connectio
```
zaheduzzaman, line 1072: s/connectio/connection

britram:
```suggestion
and as such that all packets belonging to the same active QUIC connection
```

PR #458, opened 2022-02-21
Editorial: Explain impact of reordering

mirjak:
Based Al's opsdir feedback. Thanks!

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -287,7 +287,7 @@ Hello, and the server can send 1-RTT data as soon as it has sent its Server
 Hello. The Client Completion flight contains at least one Handshake packet and
 could also include an Initial packet. QUIC packets in separate contexts during
 the handshake can be coalesced (see {{coalesced}}) in order to reduce the
-number of UDP datagrams sent during the handshake. 
+number of UDP datagrams sent during the handshake.
 
 Handshake packets can arrive out-of-order without impacting the handshake as
 long as the reordering did not cause extensive delays which would be consider
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -290,10 +290,10 @@ the handshake can be coalesced (see {{coalesced}}) in order to reduce the
 number of UDP datagrams sent during the handshake.
 
 Handshake packets can arrive out-of-order without impacting the handshake as
-long as the reordering did not cause extensive delays that trigger a
+long as the reordering was not accompanied by extensive delays that trigger a
 spurious Probe Timeout ({Section 6.2 of RFC9002}).
 If QUIC packets get lost or reordered, packets belonging
-to the same flight might not be observed in close succession in time, though
+to the same flight might not be observed in close time succession, though
 the sequence of the flights will not change, because one flight depends
 upon the peer's previous flight.
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -290,7 +290,7 @@ the handshake can be coalesced (see {{coalesced}}) in order to reduce the
 number of UDP datagrams sent during the handshake.
 
 Handshake packets can arrive out-of-order without impacting the handshake as
-long as the reordering did not cause extensive delays which would be consider
+long as the reordering did not cause extensive delays, which would be considered
 as loss by either side. If QUIC packets get lost or reordered, packets belonging
 to the same flight might not be observed in close in time, though the sequence
 of the flights will not change, because one flight depends upon the peer's
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -292,9 +292,9 @@ number of UDP datagrams sent during the handshake.
 Handshake packets can arrive out-of-order without impacting the handshake as
 long as the reordering did not cause extensive delays, which would be considered
 as loss by either side. If QUIC packets get lost or reordered, packets belonging
-to the same flight might not be observed in close in time, though the sequence
-of the flights will not change, because one flight depends upon the peer's
-previous flight.
+to the same flight might not be observed in close succession in time, though
+the sequence of the flights will not change, because one flight depends
+upon the peer's previous flight.
 
 Datagrams that contain an Initial packet (Client Initial, Server
 Initial, and some Client Completion) contain at least 1200 octets of UDP
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -287,10 +287,14 @@ Hello, and the server can send 1-RTT data as soon as it has sent its Server
 Hello. The Client Completion flight contains at least one Handshake packet and
 could also include an Initial packet. QUIC packets in separate contexts during
 the handshake can be coalesced (see {{coalesced}}) in order to reduce the
-number of UDP datagrams sent during the handshake.  QUIC packets can be lost
-and reordered, so packets within a flight might not be sent close in time,
-though the sequence of the flights will not change, because one flight depends
-upon the peer's previous flight.
+number of UDP datagrams sent during the handshake. 
+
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -287,10 +287,14 @@ Hello, and the server can send 1-RTT data as soon as it has sent its Server
 Hello. The Client Completion flight contains at least one Handshake packet and
 could also include an Initial packet. QUIC packets in separate contexts during
 the handshake can be coalesced (see {{coalesced}}) in order to reduce the
-number of UDP datagrams sent during the handshake.  QUIC packets can be lost
-and reordered, so packets within a flight might not be sent close in time,
-though the sequence of the flights will not change, because one flight depends
-upon the peer's previous flight.
+number of UDP datagrams sent during the handshake.
```
mirjak, line 290:
```suggestion
number of UDP datagrams sent during the handshake.
```

draft-ietf-quic-manageability.md
```diff
@@ -287,10 +287,14 @@ Hello, and the server can send 1-RTT data as soon as it has sent its Server
...
+
+Handshake packets can arrive out-of-order without impacting the handshake as
+long as the reordering did not cause extensive delays which would be consider
```
LPardue, line 293:
```suggestion
long as the reordering did not cause extensive delays, which would be considered
```

draft-ietf-quic-manageability.md
```diff
@@ -287,10 +287,14 @@ Hello, and the server can send 1-RTT data as soon as it has sent its Server
...
+as loss by either side. If QUIC packets get lost or reordered, packets belonging
```
ianswett, line 294: I think "did not cause extensive delays which would be consider as loss by either side." may be unclear.

In order to declare a packet lost, a later packet needs to be acknowledged, so it's more complex than time passing.  I would be inclined to remove this clause.
```suggestion
If QUIC packets get lost or reordered, packets belonging
```

ianswett (commented):
I think this might be a case where less is more.

draft-ietf-quic-manageability.md
mirjak: There are multiple cases:

1) The client never gets a reply from the server and must assume that all its initial packets got lost. I actually was briefly looking in the spec and didn't find quickly any information about a handshake retry time-out...?

2) One packet of the handshake is lost but others are acknowledged. This usually needs multiple ACK to detect loss, right?

3) Similar as tail loss, loss is detect due to a timeout. I thought that is the most likely case for loss in the handshake no?

I added this based on Al's comment in the opsdir, asking when/if reorder can impact the handshake negatively. Would be nice to say something about this, as I can definitely see who this question is interesting for network operator.

ianswett: 1. There's no handshake retry time-out in the specs.  I expect most clients have one in practice.
2. This is complex, you only need an ACK of one packet sent afterwards, but it has to be in the same PN space, so commonly you end up relying on the PTO timer.
3. Yes, this is the most likely case.  When the timer goes off, the packet is not actually declared lost, which is an important distinction, at least in my thinking.

Instead of loss, maybe "did not cause extensive delays that trigger a spurious Probe Timeout (Section 6.2 of RFC9002)."?

mirjak: Ah got it! It's about not calling it loss as it does doesn't trigger normal loss detect. Will make that edit!

draft-ietf-quic-manageability.md
```diff
@@ -287,10 +287,14 @@ Hello, and the server can send 1-RTT data as soon as it has sent its Server
...
+to the same flight might not be observed in close in time, though the sequence
```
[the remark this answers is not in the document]
mirjak: I do think I wrote this originally, so I was assuming there was a native speaker who knew what (s)he did... added "succession". Or would "observed closely in time" actually work too?

draft-ietf-quic-manageability.md
```diff
@@ -287,10 +287,14 @@ Hello, and the server can send 1-RTT data as soon as it has sent its Server
...
-though the sequence of the flights will not change, because one flight depends
-upon the peer's previous flight.
+number of UDP datagrams sent during the handshake.
+
+Handshake packets can arrive out-of-order without impacting the handshake as
+long as the reordering did not cause extensive delays, which would be considered
+as loss by either side. If QUIC packets get lost or reordered, packets belonging
+to the same flight might not be observed in close in time, though the sequence
+of the flights will not change, because one flight depends upon the peer's
+previous flight.
```
mirjak, line 297:
```suggestion
to the same flight might not be observed in close succession in time, though
the sequence of the flights will not change, because one flight depends
upon the peer's previous flight.
```

draft-ietf-quic-manageability.md
```diff
@@ -287,9 +287,13 @@ Hello, and the server can send 1-RTT data as soon as it has sent its Server
...
 could also include an Initial packet. QUIC packets in separate contexts during
 the handshake can be coalesced (see {{coalesced}}) in order to reduce the
-number of UDP datagrams sent during the handshake.  QUIC packets can be lost
-and reordered, so packets within a flight might not be sent close in time,
-though the sequence of the flights will not change, because one flight depends
+number of UDP datagrams sent during the handshake.
+
+Handshake packets can arrive out-of-order without impacting the handshake as
+long as the reordering did not cause extensive delays, which would be considered
+as loss by either side. If QUIC packets get lost or reordered, packets belonging
```
mirjak, line 294:
```suggestion
long as the reordering did not cause extensive delays that trigger a
spurious Probe Timeout ({Section 6.2 of RFC9002}).
If QUIC packets get lost or reordered, packets belonging
```

ianswett (approved):
With your suggestion, this LGTM

acmacm (changes requested):
Some edits to help...

draft-ietf-quic-manageability.md
```diff
@@ -287,9 +287,14 @@ Hello, and the server can send 1-RTT data as soon as it has sent its Server
...
-number of UDP datagrams sent during the handshake.  QUIC packets can be lost
-and reordered, so packets within a flight might not be sent close in time,
-though the sequence of the flights will not change, because one flight depends
+number of UDP datagrams sent during the handshake.
+
+Handshake packets can arrive out-of-order without impacting the handshake as
+long as the reordering did not cause extensive delays that trigger a
+spurious Probe Timeout ({Section 6.2 of RFC9002}).
+If QUIC packets get lost or reordered, packets belonging
+to the same flight might not be observed in close succession in time, though
```
acmacm, line 296: s/in close succession in time/in close time succession/

draft-ietf-quic-manageability.md
```diff
@@ -287,9 +287,14 @@ Hello, and the server can send 1-RTT data as soon as it has sent its Server
 Hello. The Client Completion flight contains at least one Handshake packet and
 could also include an Initial packet. QUIC packets in separate contexts during
 the handshake can be coalesced (see {{coalesced}}) in order to reduce the
-number of UDP datagrams sent during the handshake.  QUIC packets can be lost
-and reordered, so packets within a flight might not be sent close in time,
-though the sequence of the flights will not change, because one flight depends
+number of UDP datagrams sent during the handshake.
+
+Handshake packets can arrive out-of-order without impacting the handshake as
+long as the reordering did not cause extensive delays that trigger a
```
acmacm, line 293: s/did not cause/was not accompanied by/

mirjak:
```suggestion
long as the reordering was not accompanied by extensive delays that trigger a
```

draft-ietf-quic-manageability.md
mirjak:
```suggestion
to the same flight might not be observed in close time succession, though
```

britram (approved):

PR #459, opened 2022-03-01
Opsdir comment by Al on versions

mirjak:


britram:
lgtm

britram (approved):

draft-ietf-quic-manageability.md
```diff
@@ -518,8 +518,10 @@ number to recognize valid QUIC traffic is likely to persistently miss a fraction
 of QUIC flows and completely fail in the near future, and is therefore
 not recommended. In addition, due to the speed of evolution of the protocol,
 devices that attempt to distinguish QUIC traffic from non-QUIC traffic for
-purposes of network admission control should admit all QUIC traffic regardless
-of version.
+purposes of network admission control should not rely on the version number
+field. Instead it is recommended to admit all QUIC traffic regardless
```
acmacm, line 522: Rather than (continue to ) try to specify operator policy, and assume that the pace of operator policy changes is somehow slower than the standardization process (which it is not) replace:
      ....should not rely on the version number field. 
     Instead it is recommended to admit all QUIC traffic regardless
     of version in order to support continuous version-based evolution and
     avoid unnecessary deployment delays.
with the message from developers to operators, simply say:

```suggestion
...should not rely on the version number field alone.
Developers would prefer admission of all QUIC traffic regardless of version in order to support continuous version-based evolution. However, all parties understand the value of versions with a corresponding, fully-approved standard.
```

britram: I understand the sentiment here; I'm not sure it fits in the document, though. This looks more like an ongoing discussion than a statement of consensus.

<not a co-chair>I understand WG consensus to be that the version number field is not a useful signal within the wire image; this consensus is informed by the past few decades of deployment experience with respect to intentional and unintentional wire-image features at the transport layer. The evolvability goal of QUIC is served by the recommendation in the document as-is.</>

I'm inclined as editor to leave this as is for the next rev publication, and to continue the discussion as to whether the recommendation should stand in/around Vienna.

mirjak: Not sure if that is any better or worse:

```suggestion
field. Instead this documents recommends to admit all QUIC traffic regardless
```

mirjak: @britram do you think my PR without Al's changes is any better/clearer than the original text? If so, we could merge this PR (without Al's proposal) and open a new PR for Al's text for further discussion if needed.

However, I agree with Brian that my understanding of the group consensus is reflected in this recommendation and I don't think we can or should change that. Al's proposed text would weaken this well-discussed recommendation and therefore I don't think it would present group consensus anymore. That's my assessment but maybe the chair can also weight in @LPardue @mjoras ?

britram:
IMO the PR is clearer (and better reflects the intent / my understanding of WG consensus), so I'd merge it then continue the discussion after the submission cutoff.

PR #461, opened 2022-03-16
Document the possibility of infinite UDP reflection

### LINKED ISSUE #399: Caution that there are other protocols that amplify in packets and bytes
These have the potential of creating infinite loops when used with stateless reset.

One is ISAKMP/IKE, which conveniently does not look like a QUIC packet, but there could be others(ie: Chargegen) that are easily capable of creating infinite loops with stateless resets.

ianswett:
Fixes #399


LPardue:
Closing this PR since we closed the associated issue with no aciton

PR #467, opened 2022-04-05
Make it clear how bad a day you'll have if you switch on version

britram:
Instead of making an explicit policy recommendation with respect to version recognition, clarify the downsides of doing so.

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -537,7 +537,7 @@ and completely fail in the near future. Reliance on the version number field
 for the purposes of admission control is similarly likely to rapidly lead to
 unintended failure modes. Admission of QUIC traffic regardless of version
 avoids these failure modes, avoids unnecessary deployment delays, and
-supports continuous continious version-based evolution.
+supports continuous version-based evolution.
 
 # Network-Visible Information about QUIC Flows
 
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -529,10 +529,11 @@ information will also cause the connection to fail.
 
 QUIC is expected to evolve rapidly, so new versions, both experimental and IETF
 standard versions, will be deployed on the Internet more often than with
-traditional Internet- and transport-layer protocols. Use of the version number
-field for traffic recognition will therefore behave differently than with
-these traditional protocols. Using a particular version number to recognize
-valid QUIC traffic is likely to persistently miss a fraction of QUIC flows,
+other commonly deployed Internet- and transport-layer protocols. Use
+of the version number field for traffic recognition will therefore behave
+differently than with these protocols. Using a particular version number
+to recognize valid QUIC traffic is likely to persistently miss a fraction of
+QUIC flows,
 and completely fail in the near future. Reliance on the version number field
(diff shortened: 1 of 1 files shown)


britram:
@acmacm I think this addresses the objection to policy recommendations on version recognition; PTAL. thanks!!

draft-ietf-quic-manageability.md
```diff
@@ -529,16 +529,15 @@ information will also cause the connection to fail.
...
-
+traditional Internet- and transport-layer protocols. Use of the version number
+field for traffic recognition will therefore behave differently than with
+these traditional protocols. Using a particular version number to recognize
+valid QUIC traffic is likely to persistently miss a fraction of QUIC flows,
+and completely fail in the near future. Reliance on the version number field
+for the purposes of admission control is similarly likely to rapidly lead to
+unintended failure modes. Admission of QUIC traffic regardless of version
+avoids these failure modes, avoids unnecessary deployment delays, and
+supports continuous continious version-based evolution.
```
LPardue, line 540:
```suggestion
supports continuous version-based evolution.
```

or did you really mean "continuous continuous"?

acmacm:
I can live with the latest wording here, thanks!  And I didn't imagine that "traditional" was non-inclusive language; traditions are an inexorable aspect of many cultures...

LPardue:
I can see how some uses of the word tradition or traditional are non-inclusive. The occurences here do not strike me that way and the alternatives presented weaken the point of the text. We can proceed with an open mind to whether the term needs to be changed in response to wider review via IETF last call IMO.

mirjak:
My understanding is that "traditional" is also flagged because it's usually not a clear term in technical documents and I think that's the case here as well. There is no tradition about deploying certain protocols. We could just say "older" or "other protocols used so far" or something like this.

LPardue:
If you editors would like to replace it with a different term that conveys your intent, that's fine by me. Perhaps switch traditional for "other commonly used"

britram:
s/traditional/most other/g?

britram:
other commonly used is also good, but requires a manual rewrap. :)

draft-ietf-quic-manageability.md
```diff
@@ -529,15 +529,15 @@ information will also cause the connection to fail.
...
-not recommended. In addition, due to the speed of evolution of the protocol,
-devices that attempt to distinguish QUIC traffic from non-QUIC traffic for
-purposes of network admission control should not rely on the version number
-field. Instead it is recommended to admit all QUIC traffic regardless
-of version in order to support continuous version-based evolution and
-avoid unnecessary deployment delays.
+traditional Internet- and transport-layer protocols. Use of the version number
+field for traffic recognition will therefore behave differently than with
+these traditional protocols. Using a particular version number to recognize
+valid QUIC traffic is likely to persistently miss a fraction of QUIC flows,
```
britram, line 535:
```suggestion
other commonly deployed Internet- and transport-layer protocols. Use
of the version number field for traffic recognition will therefore behave
differently than with these protocols. Using a particular version number
to recognize valid QUIC traffic is likely to persistently miss a fraction of
QUIC flows,
```

LPardue (approved):
thanks for addressing the in-solidarity bot

PR #486, opened 2022-05-18
Inclusive language pass

### LINKED ISSUE #480: IESG Review Lars Eggert: terminology
Found terminology that should be reviewed for inclusivity; see
https://www.rfc-editor.org/part2/#inclusive_language for background and more
guidance:

 * Term "traditional"; alternatives might be "classic", "classical", "common",
   "conventional", "customary", "fixed", "habitual", "historic",
   "long-established", "popular", "prescribed", "regular", "rooted",
   "time-honored", "universal", "widely used", "widespread"
 * Term "invalid"; alternatives might be "not valid", "unenforceable", "not
   binding", "inoperative", "illegitimate", "incorrect", "improper",
   "unacceptable", "inapplicable", "revoked", "rescinded"

britram:
Fixes #480

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -605,9 +605,9 @@ A related question is whether the first packet of a given flow on a port known
 to be associated with QUIC is a valid QUIC packet.  This determination supports
 in-network filtering of garbage UDP packets (reflection attacks, random
 backscatter, etc.). While heuristics based on the first byte of the packet
-(packet type) could be used to separate valid from invalid first packet types,
-the deployment of such heuristics is not recommended, as bits in the first byte
-may have different meanings in future versions of the protocol.
+(packet type) could be used to distinguish garbage from non-garbage first packet
+types, the deployment of such heuristics is not recommended, as bits in the
+first byte may have different meanings in future versions of the protocol.
 
 ## Connection Confirmation {#sec-confirm}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -1113,7 +1113,7 @@ connection could result in variations in order, delivery rate, and drop rate.
 As feedback about loss or delay of each packet is used as input to
 the congestion controller, these variations could adversely affect performance.
 Depending on the loss recovery mechanism implemented, QUIC may be
-more tolerant of packet re-ordering than traditional TCP traffic (see
+more tolerant of packet re-ordering than TCP traffic (see
```
mirjak, line 1116: While "traditional" is not an accurate wording here, we need some adjective here because TCP with RACK might be as tolerant than QUIC.

draft-ietf-quic-manageability.md
```diff
@@ -605,9 +605,9 @@ A related question is whether the first packet of a given flow on a port known
 to be associated with QUIC is a valid QUIC packet.  This determination supports
 in-network filtering of garbage UDP packets (reflection attacks, random
 backscatter, etc.). While heuristics based on the first byte of the packet
-(packet type) could be used to separate valid from invalid first packet types,
-the deployment of such heuristics is not recommended, as bits in the first byte
-may have different meanings in future versions of the protocol.
+(packet type) could be used to distinguish garbage from non-garbage first packet
```
mirjak, line 608: why not just use non-valid?

mirjak: Actually this term just happened to be discussed between the IESG and RPC based on another case and in this context use of invalid seems fine. So maybe we just want to keep it.

PR #487, opened 2022-05-18
Note that ephemeral ports aren't usually reflectors

### LINKED ISSUE #475: IESG Review Paul Wouters: Section 8.1 - source port blocking
Section 8.1 talks about "Services might block source ports" and then lists a bunch. But contacting a service should really happen from an ephemeral source port, which would be as per [RFC 6335](https://datatracker.ietf.org/doc/rfc6335/) suggestion the range 49152–65535. None of the examples in this section come from these ports, so it feels like if one picks proper ephemeral ports, there is no relevant blocking issue at all that this section talks about. Perhaps that is the advise this section should give?

britram:
fixes #475

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -662,9 +662,9 @@ a period of time, downgrading to a non-UDP protocol (see {{fallback}}).
 
 As a result, client implementations are encouraged to avoid using source ports
 associated with protocols known to be vulnerable to reflection attacks. Note
-that the list above is not exhaustive; other source ports might be considered
-reflection vectors as well.
-
+that following the general guidance for client implementations given in
+{{?RFC6335}}, to use ephemeral ports in the range 49152–65535, has the
+effect of avoiding these ports.
 
 # Connection Migration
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-applicability.md
```diff
@@ -662,9 +662,9 @@ a period of time, downgrading to a non-UDP protocol (see {{fallback}}).
 
 As a result, client implementations are encouraged to avoid using source ports
 associated with protocols known to be vulnerable to reflection attacks. Note
-that the list above is not exhaustive; other source ports might be considered
-reflection vectors as well.
-
+that following the general guidance for client implementations given in
+{{?RFC6335}}, to use ephemeral ports in the range 49152–65535, has the
+effect of avoiding these ports.
```
mirjak, line 667: Fine to add this advice, however, this came up because in practice this advise is not always followed...

However, why did you remove the existing text rather than just adding a pointer to RFC6335?

britram, line 667:
```suggestion
that following the general guidance for client implementations given in
{{?RFC6335}}, to use ephemeral ports in the range 49152–65535, has the
effect of avoiding these ports. Note that other source ports might be
reflection vectors as well.
```

Issue #492, opened 2022-06-01
IESG Review Eric Vyncke: Section 2.1 - CID in short headers optional or not? + spin bit


mirjak:
The text appears inconsistent to me:

- "Short headers contain only an optional destination connection ID"

- "source and destination connection ID: short and long headers carry a destination connection ID", i.e., is it optional or not ?

Also some inconsistency about the spin bit, is it only in v1 or is it an invariant ?

PR #494, opened 2022-06-01
Eric Vyncke: CID field optional

mirjak:
clarify why CID is present in all headers but is also optional... see issue #492

### DIFF
--- a/draft-ietf-quic-manageability.md
+++ b/draft-ietf-quic-manageability.md
@@ -173,7 +173,8 @@ QUIC:
   (see {{Section 22.2 of QUIC-TRANSPORT}}).
 
 - source and destination connection ID: short and long headers carry a
-  destination connection ID, a variable-length field that can be used to
+  destination connection ID, a variable-length field which can also be zero-length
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-manageability.md
```diff
@@ -173,7 +173,8 @@ QUIC:
   (see {{Section 22.2 of QUIC-TRANSPORT}}).
 
 - source and destination connection ID: short and long headers carry a
-  destination connection ID, a variable-length field that can be used to
+  destination connection ID, a variable-length field which can also be zero-length
+  but if present can be used to
```
britram, line 177:
```suggestion
  destination connection ID, a variable-length field which, if of non-zero
  length, can be used to
```

britram: (not sure this is better)

mirjak: Maybe this?
```suggestion
  destination connection ID, a variable-length field which, if not zero-length,
  can be used to
```

PR #503, opened 2022-07-05
ref to draft-ietf-quic-ack-frequency

### LINKED ISSUE #473: IESG review Paul Wouters: Section 7 - ack strategy
Section 7 talks about acknowledgment strategy, but it is not actually giving any examples of what the application can do. Since it is kind of a transport layer property and not an application layer property, it is unclear how an application [developer] can replace the "TCP like ack every other packet" functionality. Are there QUIC options to consider that do this? If so, why not mention those here?

mirjak:
fixes #473

### DIFF
--- a/draft-ietf-quic-applicability.md
+++ b/draft-ietf-quic-applicability.md
@@ -592,7 +592,7 @@ types of network.
 Applications might be able to improve overall performance
 by using alternative strategies that reduce the rate of acknowledgments.
 {{?I-D.ietf-quic-ack-frequency}} describes an extension to signal the desired
-delay of acknowledgments and discusses use cases as well as impications for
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-applicability.md
```diff
@@ -591,6 +591,10 @@ contribute to link utilization, which can impact performance over some
 types of network.
 Applications might be able to improve overall performance
 by using alternative strategies that reduce the rate of acknowledgments.
+{{?I-D.ietf-quic-ack-frequency}} describes an extension to signal the desired
+delay of acknowledgments and discusses use cases as well as impications for
```
britram, line 595:
```suggestion
delay of acknowledgments and discusses use cases as well as implications for
```

britram (approved):

